PR #2, opened 2016-09-04
Allow empty AsciiDoc Attribute lists

bexelbie:
Directives, such as include:: often have empty attribute lists


mquinson:
I feel a bit bad about this one. Do you have any pointer showing that AsciiDoc attribute lists can be empty (an example would suffice).
That's a stupid question, but I never used this format myself.

Thanks, Mt.

bexelbie:
See: http://www.methods.co.nz/asciidoc/userguide.html#X21

> The list consists of zero or more positional attribute values followed by zero or more named attribute values.

See also: https://github.com/mquinson/po4a/blame/master/t/data-30/Tables.asciidoc#L34

mquinson:
Many thanks. Merged, then.

PR #5, opened 2016-09-08
Zanata fix

bexelbie:


bexelbie:
I could not get the bin/ca.po to upload, but I uploaded all of the rest.

I personally use the zanata-python-cli

bexelbie:
I also pushed a zanata.xml for the website.

I redid the translate.zanata.org project to have three projects:

* bin
* pod
* www

This is one for each gettext directory.

If we want to use zanata from now on, we should look at changing the website to reflect that.

PR #6, opened 2016-11-12
Add an option to not wrap attributes

dbarbier:
Contributed by Ineiev, thanks.


mquinson:
Hello,

a little test would have been welcome, if possible. But that's not blocking at all.

Thanks for that

mquinson:
Unfortunately, this is breaking a test! I cannot accept it as is, that would not be reasonable. Could you please fix it (either the test or the code) ?

Thanks,

PR #14, opened 2017-03-29
It seems there is still a bug on list management

jnavila:
On the attached example, the list nature of the entry is discarded.

### DIFF
--- a/t/data-30/Lists.asciidoc
+++ b/t/data-30/Lists.asciidoc
@@ -135,6 +135,16 @@ Nested list appended list item one paragraph
 
 -  Nested list item two.
 
+* Adds content from all `*.txt` files under `Documentation` directory
+and its subdirectories:
++
+------------
+$ git add Documentation/\*.txt
+------------
++
+Note that the asterisk `*` is quoted from the shell in this
+example; this lets the command include the files from
+subdirectories of `Documentation/` directory.
 
 List Block
 ----------
--- a/t/data-30/Lists.out
+++ b/t/data-30/Lists.out
@@ -133,6 +133,16 @@ Nested list appended list item one paragraph
 
 -  Nested list item two.
 
+* Adds content from all `*.txt` files under `Documentation` directory
+and its subdirectories:
++
+------------
+$ git add Documentation/\*.txt
+------------
++
+Note that the asterisk `*` is quoted from the shell in this
+example; this lets the command include the files from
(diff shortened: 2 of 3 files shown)


mquinson:
Could you please be more explicit about what is expected here? I'm not sure I see the problem...

jnavila:
Normally, the part that was added is a bullet list item (with a star character `*`) and it is rendered as such if you run `asciidoc` on the document. But if you run `po4a-normalize` on it, you can see in the `po4a-normalize.output` file that the star character has vanished in the process.

jnavila:
JFTR, the bug was discovered while reviewing this:

https://github.com/git/git/blob/master/Documentation/git-add.txt#L191

Fat-Zer:
In other words, po4a fails to parse list items improperly aligned:

e.g.:
````
// This works fine:
- List item with quite a long properly aligned wrapped
  line

// And This fails:
- List item with quite a long not-so-properly aligned wrapped
line //< notice no spaces
````

In the second case po4a just removes the list mark from the output file.

According to asciidoc userguide [1] 
> List item indentation is optional and does not determine nesting, indentation does however make the source more readable. 

So this is a valid, but discouraged syntax, and probably should be properly parsed by po4a.

@jnavila note that besides the problem in the po4a, it would be nice to fix up it in git's doc itself at least to enhance the readability of raw documentation a bit:
```
* Adds content from all `*.txt` files under `Documentation` directory
  and its subdirectories:  //< notice two spaces.
```

[1] http://www.methods.co.nz/asciidoc/chunked/ch17.html

jnavila:
@Fat-Zer That's also right. It's as if the writers for git's documentation were looking for all the discouraged and primitive corner case of the "specification".

mquinson:
I had a look at our code, and that's not easy to fix. The solution could maybe to add a warning in the asciidoc and asciidoctor tools so that the users stop using it?

mquinson:
Ok, this one is fixed, but hte code really becomes hairly. I'm not sure of how to solve this on the long term. If we don't find any more bugs, that's fine. If we do, we may need to reimplement this otherwise to make it less obscure. 

But I have no idea of how, and that's not the source code of AsciiDoc with its file of 6000 lines of python, redefining generic stuff as ordered dicts within the main file, that gives me a lot of insight. I'm affraid we're doomed of maintaining that code as is for a while...

PR #18, opened 2017-04-05
docbook: delete an obviously obsolete documentation paragraph

Fat-Zer:


mquinson:
woopsy :) Good catch, thanks.

PR #20, opened 2017-04-05
po4a.pod: minor fix to description of KernelHelp

Fat-Zer:
Current documentation makes think that KernelHelp messages of only exact =2.4.x kernels are handled
and recent versions may not, this commit changes it a bit.


Fat-Zer:
I'm feeling I propose too a lot of small changes... feel free to tell me If I'm overdoing it a bit...

mquinson:
Merging your changes is one click, so don't stop the flow of good improvements!

PR #23, opened 2017-04-10
A couple more enhancements for po4a(7)

Fat-Zer:
Note that the second patch need a review...

### DIFF
--- a/doc/po4a.7.pod
+++ b/doc/po4a.7.pod
@@ -314,8 +314,8 @@ options.
 =item -
 
 Actually translate what should be translated. For that, you have to rename
-the POT file for example to F<doc.XX.po> (where I<XX> is the ISO639 code of the
-language you are translating to, e.g. B<fr> for French), and edit the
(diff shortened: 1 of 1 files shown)


mquinson:
Hello, thanks for the patch.

I did review it, but I must confess I can't remember anymore how the addendum work, actually :( If you have a reasonnable confidence in your change, I'd trust you on this one. Do you really think I should dig into the code again, or it should be good?

Thanks for your time,
Mt

Fat-Zer:
I haven't dug the code either, but that's only reasonable way I can imagine the mode=before should work... And I can confirm that it works that way (as far as I can understand my own text=)). Also I verified that all examples work as described just now.

I've asked about a review due to I'm a bit concerned about problems with my grammar/punctuation/obscurity of formulations/whatever...

PS: just a small change to avoid confusion and make later examples produce a valid POD then just the header replaced. (add extra empty line to the end of file)

mquinson:
Ok. It sounds good, then. Merged -- thanks.

PR #24, opened 2017-04-10
Update Russion translation of the program itself

Fat-Zer:


mquinson:
Thanks for this one and all the others ;)

PR #25, opened 2017-04-11
Add keyvalue option to not wrap key:value paras

### LINKED ISSUE #8: Key/Value Pairs should be treated as verbatim
Many static site generators, amongst other things, require a set of header information at the top of the file such as:

```
Title: Some title
Date: Some date

lorem ipsum ...

foo bar baz ...
```

We should be able to optionally allow any paragraph formatted as key:value to be treated as verbatim.  I have a potential fix (needs tests) here: https://github.com/bexelbie/po4a/tree/keyvalue_text

bexelbie:
Fixes #8

Signed-off-by: Brian (bex) Exelbierd <bex@pobox.com>

### DIFF
--- a/MANIFEST
+++ b/MANIFEST
@@ -126,6 +126,7 @@ t/27-xml.t
 t/28-msguntypot.t
 t/29-wml.t
 t/30-asciidoc.t
+t/31-text.t
 t/compare-po.pl
 t/data-02/man
 t/data-02/man.fr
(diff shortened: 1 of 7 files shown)


mquinson:
Thanks for the test case, too!

bexelbie:
Any further feedback on the test case is very welcome!

mquinson:
My feeling is that Markdown should be split in another format, separated from Text. There is already enough Markdown dialects that you don't want to take all of them as options to the Text module.

But that's not a strong feeling either ;)

bexelbie:
I agree with your comment about splitting the formats.  I don't have time to work on it right now though :(

mquinson:
We should probably open a new issue to keep track of this, don't you think  ?

PR #29, opened 2017-04-14
po4a-build.xml: fix a copy-paste mistake in --version arg

Fat-Zer:


mquinson:
Thanks for being so attentive.

PR #39, opened 2017-07-25
Remove check for invalid attribute lists

bexelbie:
We allow empty attribute lists, the check does not and is unneeded


mquinson:
thanks!

PR #40, opened 2017-08-23
make wml title translatable

yangfl:


mquinson:
Hello,

many thanks for this change. Could you please add a little test case to `t/29-wml.t` and `t/data-29/general.wml` to make sure that it works and keeps doing so in the future?

Thanks in advance, 
Mt.

mquinson:
thanks

PR #50, opened 2017-09-11
Update from Weblate.

weblate:


mquinson:
Hello @rffontenelle thanks for these changes. 

Unfortunately, mode=before should not be translated. I mean, the users should write it in english in their addendum description. Could you please fix your file?

Thanks in advance, 
Mt.

rffontenelle:
Thanks for letting me know, @mquinson 

In the string

```
#. type: textblock
#: doc/po4a.7.pod:154
msgid ""
"you should put a B<position> matching this line, and a B<beginboundary> "
"matching the beginning of the next section (i.e., B<^\\.SH>). The addendum "
"will then be added B<after> the I<position point> and immediately B<before> "
"the first line matching the B<beginboundary>. That is to say:"
```

it seems the `B<before>` and `B<after>` words were inserted in a normal English sentence but also marked with B<> tag to show them as special words. What would be the approach to translate this? Should I write the sentence putting the original words in parenthesis?

Another doubt: where should I send the updated PO to: Weblate or in this Pull Request?

mquinson:
Hello @rffontenelle, thanks for your quick answer. I will accept this PR so that you can do another one easily from weblate. Please let me know if it's not convinient for you.

For the record, here is how we translated this text to french: https://github.com/mquinson/po4a/blob/master/po/pod/fr.po#L5239 
We keep the english keywords unmodified (begin, after, mode, beginboundary, etc), along with the translation between parenthesis. We sometimes even keep the english word alone when adding the translation would make the text too difficult to read.

I wish I had a way to give this warning to the translators, but I cannot find any. I find the problem in pt_BR only accidentally.

Thanks for your help,
Mt.

rffontenelle:
> Hello @rffontenelle, thanks for your quick answer. I will accept this PR so that you can do another one easily from weblate. Please let me know if it's not convinient for you.

No problem at all.

> https://github.com/mquinson/po4a/blob/master/po/pod/fr.po#L5239

Actually, the string I quoted above is further down, which in French PO file is in line 5420 (https://github.com/mquinson/po4a/blob/master/po/pod/fr.po#L5420).

In this string, looks like the `B<before>` and `B<after>` were translated. Would it be correct or I should try to keep the original (in parenthesis, if needed) ?
```
#. type: textblock
#: doc/po4a.7.pod:154
msgid ""
"you should put a B<position> matching this line, and a B<beginboundary> "
"matching the beginning of the next section (i.e., B<^\\.SH>). The addendum "
"will then be added B<after> the I<position point> and immediately B<before> "
"the first line matching the B<beginboundary>. That is to say:"
msgstr ""
"vous devez utiliser un champ B<position> correspondant à cette ligne, et un "
"champ B<beginboundary> correspondant au début de la section suivante (c'est-"
"à-dire B<^\\.SH>). L'addendum sera placé B<après> le I<point de position> et "
"immédiatement B<avant> la première ligne correspondant au champ "
"B<beginboundary>. S<C'est-à-dire :>"
```

mquinson:
Woops, you're right. Let me fix the french translation, too :) Ahem. 

The right way (IMHO) is to keep it in english with the translation between parenthesis.

rffontenelle:
Ok, thanks for clarifying. I'll fix pt_BR too.

mquinson:
Wait. Actually, here, before and after are not the keywords but the english words. It makes no sense to keep them in english. Sorry for the confusion, I'm multitasking too much these days.

My initial request was about this sentence, where I'm positive: you shouldn't translate it:
```
msgstr "Para evitar isto, você pode realizar o mesmo usando B<modo>=I<antes>:" 
```

rffontenelle:
Got it.

PR #54, opened 2017-10-04
bugfix: asciidoc: handle macros with empty target

martijnthe:
Asciidoc macros with an empty target were not matched as being a macro.
An example of a macro with an emtpy target is `endif::[]`

Signed-off-by: Martijn Thé <post@martijnthe.nl>

### DIFF
--- a/lib/Locale/Po4a/AsciiDoc.pm
+++ b/lib/Locale/Po4a/AsciiDoc.pm
@@ -542,7 +542,7 @@ sub parse {
             }
             @comments=();
         } elsif (not defined $self->{verbatim} and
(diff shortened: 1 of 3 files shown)


martijnthe:
@mquinson: I noticed this message on the po4a website:

> You can also submit Pull Requests against this GitHub repository, even if we try to reduce our dependency on GitHub to the minimum ourselves.

Where is the official / upstream git repo? How do I get this PR into the official upstream?

mquinson:
Hello,

thanks for the change. GitHub is the easiest way to integrate your changes right now. We are using debian's Alioth as a main development site, but alioth will be turned off at some point, so GitHub is just fine until we find a new better home.

Thanks again for this change, 
Mt.

martijnthe:
@mquinson thanks!

PR #59, opened 2017-10-29
Fix xml closing tags in pt translations

jnavila:


mquinson:
Thanks for this, Jean-Noël

PR #62, opened 2017-11-12
Fix non-ATX titles in AsciiDoc

bexelbie:
The subtraction of the new line is no longer needed


mquinson:
Hello,

this breaks the tests. It seems that you forgot to add the po files of the test into the PR. 

Thanks,

bexelbie:
So, as far as I can tell, the current HEAD fails at the AB titles tests on my machine.  This patch passes on my machine.  The .po files didn't change.  I am starting to think there is something different about my build environment.

bexelbie:
Ok - I found it.  It is my machine - there is a missing dependency that isn't being flagged :(

Issue #61, opened 2017-11-11
AsciiDoc tables not processed correctly


bexelbie:
Tables are being word wrapped which changes their layout post-translation.

I believe the right answer is to mark tables as no-wrap

PR #63, opened 2017-11-12
Make Tables a no-wrap output in AsciiDoc

bexelbie:
Table formats have changed.  The old formats are deprecated
and parsed by neither asciidoc or asciidoctor.

Treating tables as no-wrap passes them as preserved

Tentative fix for #61 - do not merge yet

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -10,6 +10,7 @@ __   __/ _ \ | ___| ___|
 
 AsciiDoc:
  * Fix Asciidoc unindented lists (GitHub's #149)
+ * Make Asciidoc Tables nowrap and support current format (GitHub's #63)
 
 Translations:
  * Updated: German, thanks Helge Kreutzmann.
--- a/lib/Locale/Po4a/AsciiDoc.pm
+++ b/lib/Locale/Po4a/AsciiDoc.pm
@@ -325,7 +325,6 @@ sub parse {
                  ($paragraph =~ m/^[^\n]*\n$/s) and
                  # subtract one because chars includes the newline on the paragraph
                  ((chars($paragraph, $self->{TT}{po_in}{encoder}, $ref) - 1) == (length($line)))) {
-
             # Found title
             $wrapped_mode = 0;
             my $level = $line;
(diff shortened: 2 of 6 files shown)


mquinson:
Hello, thanks for this. I was about to merge it when I saw your request to not.

I am willing to merge any PR coming from you because of your previous track, but I must confess that I'm not convinced by the amount of tests that you have for this new feature ;)

bexelbie:
I appreciate your confidence :).

I requested this not get merged for two reasons:

1. I have a localizer who is testing the output to see if it makes sense to them.

2. I only converted the existing test set.  I had a feeling you'd like to see more :P.  Any ideas what else you'd like to see?  I can throw the examples from the asciidoctor/asciidoc sites in.

mquinson:
Well, I think that we need at least an example that would wrap. The
existing one may wrap, but I'd prefer to have a really large table to
that extend.

Also, it seems to me that you changed the test to account for another
syntax to declare tables, unless I'm mistaken. If so, it'd be good to
have tests for all variants, don't you think?

A good practice is to add as a test for each thing that breaks in the
document you're handling.

Thanks for all this,
Mt.

bexelbie:
I'll look for a larger table to put in the testing.

The variants I have breaking are all included.

The syntax in the test is deprecated and as far as I can tell is no longer supported by either parser.

bexelbie:
I've updated the tables tests with what I believe you were looking for.  A larger table and some different spacing styles.

I've got it out to a translator for review as well.

mquinson:
Hello @bexelbie,

could you please remind me what is still missing before we can merge this PR?

Thanks, 
Mt

mquinson:
Hello @bexelbie, as I feared, this PR went out of sync with the code. I'd love if it could be cleaned up and merged so that we can forget about it...

Thanks for your help,

bexelbie:
I embarrassingly forgot about this PR.  Thanks for the ping.  I am rebasing it now.

mquinson:
Thanks @bexelbie. Does it mean that you feel like this PR is now ready to be merged? You said earlier that you'd like more test and more eyeballs on the output. 

I'm still embarrassed to propose more tests, as I prefer doing white-box testing and I don't know the code. I was hoping that you could reuse some chunks of the projects on which you are using po4a in production.

bexelbie:
This patch includes tests in test framework.  The new tests are drawn from my production use cases.  I do believe this is ready to be merged.  All AsciiDoc tests seemed to pass in my testing.  I removed the "todo" from Tables, if I did it correctly.

mquinson:
Perfect, then. Merged. Many thanks for this!

PR #66, opened 2017-11-19
Add YAML file support

bexelbie:
This implements YAML parsing with YAML::Tiny

### DIFF
--- a/README
+++ b/README
@@ -24,6 +24,7 @@ module for the following formats:
   - wml: WML documents.
   - xhtml: XHTML documents.
   - xml: generic XML documents (see also docbook).
+  - yaml: YAML documents.
 
 INSTALLATION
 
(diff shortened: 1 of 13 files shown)


bexelbie:
I'd like to not merge this yet either, but I did want to submit it for feedback.

It adds a dependency on YAML::Tiny.  I suspect it is packaged for Debian.

mquinson:
Hello,

why exactly you don't want to get this merged yet? It sounds very good to me, don't you think? I do not see what to add...

bexelbie:
I was hoping to beat it up a bit more.  At this point, let's merge it and I'll patch it if I find anythign down the road.

mquinson:
Yes, exactly. You have a decent code, a decent documentation, some decent tests, so you're more than ready for the big day. Let's fix the bug when they appear afterward (if they do) instead of staring at the code to see eventual future issues :)

PR #79, opened 2018-02-10
Update from Weblate.

weblate:


bexelbie:
@mquinson What's the process from these?  Do we immediately update hte site after these are commited?  We should write a guide so we all do it the same way.

mquinson:
Like I said, yes, that's a good idea. Plus, I usually check that the change is mentioned in the NEWS file.

PR #113, opened 2018-05-24
Keep the old translation files if the keep threshold isn't met.

bfg1981:
Prevent the translated files from being remove if the keep threshold isn't met.

### DIFF
--- a/po4a
+++ b/po4a
@@ -390,7 +390,9 @@ Do not change the POT and PO files, only the translation may be updated.
 
 =item B<--no-rm-translations>
 
-Keeps the old translation files if the translation doesn't meet the threshold specified by --keep.
+Keeps the old translation files if the translation doesn't meet the threshold specified by --keep. This is useful in the case where you have existing translations, but the sources gets updated and you fall below your desired threshold.
+
+WARNING: This may result in very high grade translations of outdated files.
 
 =item B<--rm-translations>
 
@@ -1498,19 +1500,20 @@ if (not $po4a_opts{"no-translations"}) {
             $doc->{TT}{po_in} = $po;
             $doc->{TT}{po_in}->stats_clear();
 
-	    my $translationBackupfile;
+	    my $translation_out;
 	    if ($po4a_opts{"no-rm-translations"} &&
 		-e $document{$master}{$lang}) {
-		(undef,$translationBackupfile)=
-		    File::Temp::tempfile("po4a-transBackupXXXX",
-					 DIR    => $ENV{TMPDIR} || "/tmp",
+	      (undef,$translation_out) =
+		    File::Temp::tempfile("$document{$master}{$lang}.new-XXXX",
 					 OPEN   => 0,
 					 UNLINK => 0);
-		rename ($document{$master}{$lang}, $translationBackupfile);
+	    }
+	    else {
+	      $translation_out=$document{$master}{$lang};
 	    }
 
             $doc->process('file_in_name'  => \@file_in_name,
-                          'file_out_name' => $document{$master}{$lang},
+                          'file_out_name' => $translation_out,
                           'file_in_charset'  => $file_opts{"mastchar"},
                           'file_out_charset' => $file_opts{"locchar"},
                           'addendum_charset' => $file_opts{"addchar"},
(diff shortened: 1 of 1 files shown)


mquinson:
Hello,

many thanks for this patch. Could you please elaborate on the advantage of this solution compared to changing the threshold to 0? I usually achieve the same by passing ```-k 0``` as an argument. But maybe I missed something in your change?

Thanks again,
Mt.

bfg1981:
It's quite the opposite. My use case right now is set -k 0.95 and --no-rm-translations. Here's the scenario.

1. We have an existing target translation at >95%
2. Someone updates the source files, so the translation drops below the threshold.
3a. Without the --no-rm-translations the existing target translation is deleted.
3b. With the --no-rm-translations the outdated target translation is kept
4. If the threshold is once again passed, the target translations are updated in both scenarios.

mquinson:
And what is the difference between ```-k 0.95 --no-rm-translations``` and ```-k 0``` scenarios?

thx

bfg1981:
In the -k 0 scenario there's no effect from the --no-rm-translations as both alternatives become the same. The target translations are always generated, but with potentially very low translation grade. For us it's not desirable to generate a translation which is potentially in our case mostly English and not the target language.

mquinson:
So... you mean that with your change, the translations are not removed if they pass below the threshold, but they are not created in the first place if they never met that threshold, am I right?

bfg1981:
You are right!

mquinson:
I now better understand your change, and I agree that this is useful, thanks.

I will review the code, as I have some comments here.

ciampix:
As this is not trivial, it could be useful to create a flowchart to better explain the entire process... how about it?
I know "use the source Luke, etc." but "a picture is worth a thousand words" too... :-)

po4a
```diff
@@ -1492,6 +1498,17 @@ if (not $po4a_opts{"no-translations"}) {
             $doc->{TT}{po_in} = $po;
             $doc->{TT}{po_in}->stats_clear();
 
+	    my $translationBackupfile;
+	    if ($po4a_opts{"no-rm-translations"} &&
+		-e $document{$master}{$lang}) {
+		(undef,$translationBackupfile)=
+		    File::Temp::tempfile("po4a-transBackupXXXX",
```
mquinson: Could you use a better name, something like $existingName.bak or something?

po4a
```diff
@@ -1492,6 +1498,17 @@ if (not $po4a_opts{"no-translations"}) {
             $doc->{TT}{po_in} = $po;
             $doc->{TT}{po_in}->stats_clear();
 
+	    my $translationBackupfile;
```
mquinson: The name of this variable is very different from the other ones. Not a big deal but if you have a better idea, it'd be welcome.

po4a
```diff
@@ -388,6 +388,10 @@ Do not generate the translated documents, only update the POT and PO files.
 
 Do not change the POT and PO files, only the translation may be updated.
 
+=item B<--no-rm-translations>
+
+Keeps the old translation files if the translation doesn't meet the threshold specified by --keep.
```
mquinson: You should better explain the effect of this option: look how long it took me to understand :)

Plus, you should warn the users that if they activate this option, they may end up with well translated but outdated translation files. That's very different from the usual behavior of po4a and should be stressed in the doc.

po4a
bfg1981: That was actually what I did originally, but that is not production grade. If you mess with files in place you risk that someone is already using that name, better to avoid it. The tempfilename however is not something I'm proud of.

po4a
bfg1981: simply translation_backup?

po4a
bfg1981: There's nothing like 100% understanding your own thoughts and being surprised when others don't do the same. Having experienced our discussion I was expecting this comment :-)

po4a
mquinson: But unfortunately, with your schema, you loose the connexion between temp file and master file it comes from. Could we use File::Temp::tempfile("${masterfile}.bak-XXXX" maybe?

po4a
mquinson: great!

bfg1981:
I've reorganized the logic a bit so that I stage an outputfile. This is better practice as all timestamps etc. are kept from the old file. I've also changed the variable names and the tempfilename. If this is OK, I can rebase the commit to make it a single commit.

mquinson:
This is very good, thank you. I'll merge it. 

I think I'll rename the option to --keep-translations or something to avoid confusion with --rm-translations that is not as close as it seems to your proposal, but that's another story.

PR #117, opened 2018-05-26
yaml: Add references

l2dy:
This PR adds references from the yaml hierarchy.

I'll fix the tests if it passes review.

### DIFF
--- a/lib/Locale/Po4a/Yaml.pm
+++ b/lib/Locale/Po4a/Yaml.pm
@@ -64,7 +64,7 @@ sub parse_file {
         || die "Couldn't read YAML file $filename : $!";
 
     for my $i (0 .. $#{$yaml}) {
-        &walk_yaml($self, $yaml->[$i]);
+        &walk_yaml($self, $yaml->[$i], "");
     }
     $self->pushline(Encode::encode_utf8($yaml->write_string()));
 }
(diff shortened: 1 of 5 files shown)


mquinson:
Hello. This seems pretty good, but I definitely need more tests to see what it really does. It's hard to tell as is.

Many thanks in advance,
Mt

mquinson:
just to make it clear: i'd really appreciate if you could fix the existing tests, but also add some new tests that explore the intended behavior of your code.

Thx

l2dy:
I have updated the existing tests and I believe they already explore the intended behavior of this change.

mquinson:
thanks, that's in :)

l2dy:
Hi @mquinson,

Do you think `>` would be better than `:` since no line numbers are involved here?

mquinson:
Hello @l2dy. I was wondering whether ```/``` would better denote the hierarchy idea. Not such a big deal, but you are free to change it if you prefer it this way.

More concerning is the fact that the reference is sometimes dupplicated:
```
#. type: Hash Value - Key: Name
#: :Topics::Topics: :Topics::Topics:
#, no-wrap
msgid "Troubleshooting"
msgstr ""
```

Is it really the expected behavior? 

Thx

l2dy:
YAML::Tiny doesn't preserve ordering, so we can't rely on filenames and line numbers which are mostly unique for each entry.

But what happens if GNU gettext's xgettext(1) sees multiple strings on the same line? The references are deduplicated[1]! So IMHO, duplication is a problem shared across formats and deduplication shouldn't be done in the format modules.

Now, I don't have the Perl skills required to implement deduplication, but I'd say that the yaml module can stay as is regarding this. Mini test case in [2] for anyone willing to implementing deduplication.

[1]
```
  printf("\n%s\n", _("Compile options:"));printf("\n%s\n", _("Compile options:"));
```

```
#: version.c:404
msgid "Compile options:"
msgstr ""
```

[2]
```
a:
  - a
  - a
  - a
```
```
#. type: Array Element
#: :a :a :a
#, no-wrap
msgid "a"
msgstr ""
```

mquinson:
Ah, I think I got it.  In my previous example, the string "Troubleshooting" appears in two differing contexts, that happen to have the same reference in the tree. Am I right?

If so, we can definitely live with it. Thanks for your help.

l2dy:
I'm not sure why this module treat hash keys as part of the type, but at least for arrays they share the same reference (YAML::Tiny doesn't preserve ordering).

mquinson:
I just implemented the reference deduplication in po4a (my plane is in 3 hours, I'm bored), but this does not change anything in yaml example files. I'm thus suspecting that the problem is elsewhere.

Not a big deal, but it's surprising, isn't it?

l2dy:
> this does not change anything in yaml example files

No, it did. I'll make a PR soon.

PR #118, opened 2018-05-26
yaml: Use ">" instead in references

l2dy:
`//` is interpreted as `/` in UNIX path, so I'd prefer using `>` which doesn't have this convention.

```
-
  -
    - a
-
  -
    - b
```
```
#. type: Array Element
#: >>
#, no-wrap
msgid "a"
msgstr ""

#. type: Array Element
#: >>
#, no-wrap
msgid "b"
msgstr ""
```


l2dy:
Thanks!

mquinson:
thanks to YOU :)

PR #119, opened 2018-05-26
yaml: Test references

l2dy:


mquinson:
Excellent! Good catch.

PR #125, opened 2018-06-04
Fix for fortunes parsing

toddy15:
Hi Martin,

this should fix Debian bug #900649.

Regards,
Tobias


mquinson:
This is an excellent change set, thank you for your care.

PR #126, opened 2018-06-05
Code cleanup for fortunes parsing

toddy15:
Hi Martin,

I've reorganized the code for the fortunes parsing, it still passes :-) You might want to merge this as well.

As I'm fairly new to the po4a codebase: Is there a method to detect the end of file? Or is it the only way as I've done it in the code to shift the next line and see if it's defined?

Regards,
Tobias

### DIFF
--- a/lib/Locale/Po4a/Text.pm
+++ b/lib/Locale/Po4a/Text.pm
@@ -337,29 +337,34 @@ sub parse_fortunes {
     # because the formatting of the fortunes
     # is usually hand-crafted and matters.
     $wrapped_mode = 0;
-    if ($line =~ m/^%%?\s*$/) {
-        # Found end of fortune
+    # Check if there are more lines in the file.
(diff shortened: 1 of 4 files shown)


mquinson:
Hello Tobbias,

I must confess that I totally forgot :) I wrote this code almost 20 years ago, forgot about it for 10 years and I'm mostly in maintainance mode since I'm back, 3 years ago.

It seems that the other formats are using something like
```
    ($line,$ref)=$self->shiftline();

    while (defined($line)) {
................
        # Reinit the loop
        ($line,$ref)=$self->shiftline();
        undef $self->{type};
    }
```

That's ugly, but at the same time, each formats would have to implement eof() in a specific manner if it existed, and they'd be the only user of that function. So, each format can to it for itself, if it sees it fit.

mquinson:
Thanks for the improvements, by the way.

PR #127, opened 2018-06-08
Mention changed handling of fortunes

toddy15:
Hi Martin,

I'm not sure if the change is important enough to be mentioned in the NEWS. If you think it should be mentioned, this would be my suggestion.

Regards,
Tobias


mquinson:
I surely agree. Sorry for not doing so myself before.

PR #128, opened 2018-06-08
 Add support for e-mail macros .MT and .ME

toddy15:


mquinson:
As usual, this is a very high quality commit. Thanks for that. Could you please next time include a little item in the NEWS file too? Thx

PR #134, opened 2018-06-17
Update from Weblate.

weblate:


mquinson:
Maybe we can wait a bit before merging weblate PRs to reduce their amount?

PR #135, opened 2018-06-21
Spelling: "I.E", "which decay"

comradekingu:


po4a
```diff
@@ -392,7 +392,7 @@ Do not change the POT and PO files, only the translation may be updated.
 
 Keeps the existing translation files even if the translation doesn't meet the threshold specified by --keep.
 This option does not create new translation files with few content,
```
comradekingu: For some odd reason there are two whitespaces here.

mquinson: The only reason is my mistake, I guess. I'll fix it.

mquinson: I fail to see the two whitespaces, sorry. Feel free to submit another PR.

comradekingu: You see them in the diff, but not anywhere else. Possibly zero-width spaces?

PR #145, opened 2018-09-04
Ensure one line per item of array

### LINKED ISSUE #76: The Xml module does not normalize the document, making it difficult to use addenda
Initially reported on Alioth by me, back in 2005:
```
For the addenda to work, it is better to have short lines (so that the 
translator can choose between which ones to put there addenda). 
The Sgml module does so since the bug #263300, in Debian.
```

mquinson:
Attempt to fix #76, thanks to Osamu Aoki.

Signed-off-by: Osamu Aoki <osamu@debian.org>

### DIFF
--- a/lib/Locale/Po4a/TransTractor.pm
+++ b/lib/Locale/Po4a/TransTractor.pm
@@ -649,6 +649,9 @@ sub addendum {
             $self->get_out_charset);
     }
 
+    # Ensure each item of array contain "\n" terminated 1 line
+    @{$self->{TT}{doc_out}} = map { "$_\n" } map { split '\n', $_ } @{$self->{TT}{doc_out}};
+
     my $found = scalar grep { /$position/ } @{$self->{TT}{doc_out}};
     if ($found == 0) {
         warn wrap_msg(dgettext("po4a",


mquinson:
I'm creating this pull request to test the commit on travis and appveyor.

mquinson:
It is too late for me to dig into the problem tonight, I'm deeply sorry about it. @osamuaoki would you please be kind enough to check on this problem? Maybe, that's a false positive of the testing infrastructure, I have no idea.

Anyway, many thanks for your involvement and your help to po4a.

osamuaoki:
Hi

The title seems root cause fix to the problem than my quick work around.

I will look into this in a week when I find time with Debian box

Sent from iPhone 

2018/09/05 5:36、Martin Quinson <notifications@github.com>のメール:

osamuaoki:
On Tue, Sep 04, 2018 at 01:36:55PM -0700, Martin Quinson wrote:
> It is too late for me to dig into the problem tonight, I'm deeply sorry about
> it. @osamuaoki would you please be kind enough to check on this problem? Maybe,
> that's a false positive of the testing infrastructure, I have no idea.
> 
> Anyway, many thanks for your involvement and your help to po4a.

I see problem happens with addendum with upstream line ending with empty
line since it get dropped.

This is expected since perl split has strange feature of doing so.

Instead of using my brute force work-around code patch:
  @{$self->{TT}{doc_out}} = map { "$_\n" } map { split '\n', $_ } @{$self->{TT}{doc_out}};
if we preserve empty lines at the end with
  @{$self->{TT}{doc_out}} = map { "$_\n" } map { split '\n', $_, -1 } @{$self->{TT}{doc_out}};
the reported failures may be avoided.   (Needs to be checked)

But this seemingly correct rule created too many blank lines in XML for
me.   I guess I need to fix this extra line problem in XML related code.

I think the best solution is one which don't force to split at the place
I tried.  If we dare to do it, it should be done with map { split '\n',
$_, -1 }.  

Question is is there any non-XML case in which we need to run split to
get it right.  If it is wide spred problem, we keep split here with
better style.  Otherwise, we better move this split action to XML
specific data path.

Osamu

osamuaoki:
I now think workaround was not addressing the real problem.  Real problem is handling of bookinfo which contain author and copyright data.

In this debug, I enhanced debug output for easy reading.  Also, updated po4a(7) manual for ease of understanding.

Here is a patch solving the problem adding translator to the DocBook XML data from its root cause.
[po4a-patch.zip](https://github.com/mquinson/po4a/files/2363619/po4a-patch.zip)

This is not aggressive work around.

mquinson:
Closing this PR as @osamuaoki proposed a better patch. Sorry for the delay.

PR #147, opened 2018-09-24
Fix docbook bookinfo

mquinson:
New version of the PR #145

PR #148, opened 2018-09-30
[nomerge] enhance debug output, fix docbook behavior, allow separate input for POT

osamuaoki:
Sorry for my old patch containing version.

Here is my tentative work to address several issues. docbook behavior change is based on tarball test posted to https://github.com/mquinson/po4a/issues/76 .

These are important to allow translator data addition.

With this debug outputs, we see what are we matching against for addendum.  Also actual tag property setting can be seen.

I cherry picked new changes in master so it is cleanly merged.

TODO:

- Separate POT input via pot_in needs test case for po4a/po4a.conf.
- My tarball test needs to be adopted into test framework


mquinson:
Could you consider adding the tests to the pull request directly, please?

Thx

osamuaoki:
Changing tag type is certainly a workaround.  But I think that is not the safest and robust way to address addendum problem.  Give me some time to learn test code writing ...  No need to merge yet

mquinson:
Hello @osamuaoki , please take your time, your work is really welcome.

osamuaoki:
Hi,

I finally got test case for DocBook added.
Addendum issue is fixed without problem by adding one line ;-)

@{$self->{TT}{doc_out}} = map { split /(?<=\n)/, $_ } @{$self->{TT}{doc_out}};

Although, tweaking tag behavior solved issue for my upload of Debian History, this is not the approach we want to use for releasing po4a since it may cause other regression.  

So I decided Document work around in manpage.

Also, tracing code was not easy since debug output was not organized.  I have added extensive debug output for easier future debug.
This solves issues discussed at:
 * https://github.com/mquinson/po4a/pull/147 (old incorrect one-liner fix)
 * https://github.com/mquinson/po4a/issues/76

osamuaoki:
I think code is practically ready to be merged.  One thing you may want to check is I left generated file in the t/t-02-addendums directory for easy review.  But if you want cleaner build tree, move them to tmp/.

That's all.

osamuaoki:
I cleaned up history for easier review.

(i know its bad idea usually but ... it was ugly)

PR #149, opened 2018-10-02
Fix Asciidoc unindented lists

bexelbie:
Lists are not required to be indented.  This fixes them to put separate items in the .pot file


mquinson:
This looks great, thank you. Merged, and documented in NEWS file. Could you do that later point directly in the PR next time, please?

Thanks again.

PR #150, opened 2018-10-03
Add PR section, including NEWS note

bexelbie:
This adds a note about the NEWS file I keep forgetting about :)


mquinson:
Good idea, thanks.

PR #163, opened 2018-12-01
asciidoc: do not include table fences in pot

jnavila:
This is the previous behavior and it should be preserved at least.


ciampix:
AppVeyor never succeeded. Why not fix it or disable it?

mquinson:
> AppVeyor never succeeded. Why not fix it or disable it?

Any help would be welcome. I did not succeed to fix it, and I'd be ok disabling it if it remains so... 

Thanks, Mt.

PR #164, opened 2018-12-01
asciidoc: add option to prevent translation of image targets

jnavila:
Most of the time, the images are relative to the asciidoc files and
don't need special processing. This option does not change the default
behavior, so that existing projects are not broken.


mquinson:
Hello Jean-Noël, thanks for these good changes. Would you mind adding some tests to this feature, please?

Thanks in advance,

jnavila:
It seems a rebase is needed.

jnavila:
I thought the previous change to table management was not yet tagged and that this PR only brought it back to the anterior behavior.

PR #166, opened 2018-12-08
[asciidoc] add basic table cells support

jnavila:
For test and comment :+1: 

This PR introduces the management of table cells in asciidoc. The support for now is very basic and has some known limitations:

 * escaped  | are not managed. I couldn't come up with a correct regexp which would reject them. Maybe they can be managed later.
 * for cells that contain asciidoc block formatting ("a|" cells), the content is processed as a verbatim block. The general rule is that if a cell content spans on two lines or more, the text is treated as verbatim.

Both limitations may be released in the future. They are less prioritary because making complex stuff in tables is generally discouraged or thrown upon: this usually a smell that a table is used for formatting purpose and not for presentation of tabular data.

This behavior breaks the previous one, so it is introduced as an option `tablecells` of asciidoc plugin.

**TODO**: if accepted, 
 * add an entry in NEWS
 * document the option

### DIFF
--- a/lib/Locale/Po4a/AsciiDoc.pm
+++ b/lib/Locale/Po4a/AsciiDoc.pm
@@ -133,6 +133,7 @@ sub initialize {
     $self->{options}{'style'}='';
     $self->{options}{'definitions'}='';
     $self->{options}{'noimagetargets'} = 0;
+    $self->{options}{'tablecells'}=0;
 
     foreach my $opt (keys %options) {
         die wrap_mod("po4a::asciidoc",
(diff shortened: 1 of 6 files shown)


lib/Locale/Po4a/AsciiDoc.pm
```diff
@@ -128,6 +128,7 @@ sub initialize {
     $self->{options}{'macro'}='';
     $self->{options}{'style'}='';
     $self->{options}{'definitions'}='';
+    $self->{options}{'tablecells'}=0;
```
mquinson: Could you please document this option at the top of the file?

Thanks at lot,

mquinson:
The release is done, so I'm in good shape to include your PR once it's (rebased and) documented.

Thanks again,
Mt

jnavila:
Now, it works correctly :tada:

mquinson:
This is very good, many thanks. My only remark is that you could add a tiny bit more of comments in the hard part of your code (eg, near the huge split command followed by a variable capturing the result of a regex), but that's not too ugly either. 

Thanks again,
Mt

jnavila:
I'd be glad to change this part: I couldn't find a simple way to split the lines into interlaced cell contents and cell separators. Please note, that there is no submatch in the regexp. For the remaining, it's quite straightforward.

PR #177, opened 2019-06-18
Update from Weblate

weblate:


ciampix:
CI AppVeyor seems to be broken. Either we should disable it or fix it, thanks...

PR #203, opened 2020-04-07
po4a: Use value of --master-language for the language of the generated POT file

eighthave:
This lets the `Language:` header be set in the .pot file to declare what the source string language is.  This is not required nor apparently common, but it doesn't break how _gettext_ works: https://lists.gnu.org/archive/html/bug-gettext/2020-04/msg00004.html


mquinson:
Thanks a lot

PR #208, opened 2020-04-14
add gettext flag "markdown-text" for relevant entries

eighthave:
_gettext_ defines many flags for various entries that need special handling: https://www.gnu.org/software/gettext/manual/html_node/PO-Files.html  Markdown text is such a thing.  Translation tools like Weblate will automatically enable certain checks and displays based on the _gettext_ flags.  This change defines a new flag `markdown-text` to make all text entries that can have Markdown syntax in them.

(FYI, Weblate currently uses `md-text` but we're adding `markdown-text` since much less ambiguous.  The built-in _gettext_ flags do not use abbreviations.  Follow that here https://github.com/WeblateOrg/weblate/issues/3648)

### DIFF
--- a/lib/Locale/Po4a/Po.pm
+++ b/lib/Locale/Po4a/Po.pm
@@ -137,7 +137,7 @@ use Time::Local;
 use Encode;
 use Config;
 
-my @known_flags = qw(wrap no-wrap c-format fuzzy);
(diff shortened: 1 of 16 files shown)


mquinson:
Thanks a lot!

PR #224, opened 2020-05-06
Fix tests that use msginit

hattya:
The value of 'Project-Id-Version' depends on the root directory of po4a.


mquinson:
Thanks a lot!

PR #247, opened 2020-06-03
Asciidoc: force unwrapping lines

### LINKED ISSUE #242: Asciidoc: add an option to prevent line wrapping
If a paragraph of a list item contains for instance "see point A. Bla bla bla" and the wrapping occurs like that:

```
  see point 
  A. Bla bla bla
```

Then the second line can be interpreted as a sub-list start. Disabling wrapping would ensure that such occurence does not happen.

If asciidoc output is aimed at being read directly, wrapping might still be an option. But for machine processing, it's safer to disable it.

jnavila:
This option disables line wrapping when it would normally occur in
order to prevent introduction of asciidoc formatting errors due to not
managed new line prefix that could be wrongly interpreted by asciidoc.

Fixes #242


jnavila:
@mquinson What's your thought on this PR?

mquinson:
Hello,

I'm wondering whether we should call it wrap-ctn=no to align with the wrap-po option. I'm not completely sure of how to make it nice and consistent for our users. This wrap-ctn option could still be specific to the module, or more generic. It should only accept "no" or an integer as a value. Passing "newline" probably does not make any sense for the content, as msgfmt will reflow it anyway. And if we implement it this way, we should modify the text format that provides the "neverwrap" option for the same purpose.

Even if we don't implement a full wrap-ctn option, we should use the same name for the text and asciidoc plugins, eg probably "neverwrap" instead of "nowrap" right now.

EDIT: actually, the option is recognized as "neverwrap" but documented as "nowrap" both in the POD doc and in the NEWS changelog. Could you please fix it? That's probably sufficient for this PR to get merged, even if the wrap-ctn thing would be nice one day.

jnavila:
I had the issue that what we are talking about is "not wrapping the output content"when wrapping could be allowed.

I didn't think of having a common name for all plugins or even being able to specify the width of wrapping. It may make sense for other output plugins. Changing to `wrap-ctn` or similar is not a big deal and can be done in this PR. `wrap-ctn` seems not clear, the `ctn` part not really being a commonly used contraction; why not write it in full: `wrap-content`?

mquinson:
You are right. Either `wrap-content=no` or `neverwrap` are better names than my initial proposal. Also, instead of removing the wrapping of the paragraph afterward, it's probably better to use the `"wrap"` optional parameter of the `TransTractor::translate()` method. See eg https://github.com/mquinson/po4a/blob/master/lib/Locale/Po4a/Text.pm#L604

jnavila:
I don't think using the "`wrap`" option of translate is the correct way, because this option preserves the line wraps of the input text which is not the requested behavior. We still want to unwrap paragraphs when generating the po file. But when going back, we don't want to wrap them again.

mquinson:
Hello, sorry for the delay, I took some days of vacations.

I have a strong concern about this PR. The option it adds is somehow not optional. If you don't pass it, things may break in unpredictable ways, and it's really difficult as a maintainer to check that the translation rendering is OK, while it's really not easy for the translators to see the rendered document. This is why I say that this option should always be passed.

I'm not sure about what's the right solution here. Maybe we should make it a non-option, and ensure that the translation is never wrapped by po4a to prevent any potential issue. The only issue I see with this is that the diffs to the translations may be harder to read. But since translations should not be considered as source files (the PO files and the master files are the source), this issue is not a big deal.

Or maybe we could try to detect when the wrapping breaks things and try to mitigate the issue. That's not trivial to implement. We could have a regexp detecting the itemize. When a wrapped translated document matches this regexp, then we know that the issue occurred. In this case we could either try to rewrap with another column, or unwrap that specific paragraph, or even do something cleaver such as rewrapping one word only to break the issue. This approach seems hard to implement and fragile to use.

So I would tend to prefer the first solution, that has the benefit of being almost written already.

What do others think of it?

Thanks for your patience,
Mt

jnavila:
From your reply, I understand that keeping the behaviour by default is not high priority on your list. I'm ok with switching the default to "no wrapping".

mquinson:
Thanks @jnavila.

Could you please consider changing this PR (or maybe doing another one) to implement the new "unwrap all paragraphs" behavior? That would be great, and I would probably release v0.60 after that.

Thanks in advance,

jnavila:
Sorry for the very long delay. I'll try to revive this PR into what was discussed.

jnavila:
Just to be clear : that requires to possibly change all asciidoc tests in case wrapping appears.

jnavila:
Reviewing the tests, I came across a mistake of the same kind.

[This line](https://github.com/mquinson/po4a/blame/master/t/fmt/asciidoc/Lists.adoc#L135) is already showing that line wrapping can be tricky. The year at the start of the new line is interpreted as a numbered sub-list. See in po file:

https://github.com/mquinson/po4a/blob/b78b8fdb46b82e487399fc5bc0be18e873ec78b1/t/fmt/asciidoc/Lists.po#L422-L425

jnavila:
@mquinson Ready for review!

mquinson:
Thanks for these changes, @jnavila, they are really improving the things.

One last thing before we release it, please. Could you extend a bit the documentation of forcewrap to (1) say that only non-verbatim texts will be wrapped (2) explain why it's a bad idea to use this option, maybe with the tiny example you found in our test suite that clearly shows that blindly wrapping translations will lead to chaos and mayhem. 

Many thanks in advance,
Mt

mquinson:
Many thanks, that's great.

ciampix:
Sorry being late, just to say that you all are doing a great job here!

PeterNjeim:
Hello guys, sorry for necroing this but I can't seem to have no wrapping on my generated markdown files and it's breaking the syntax. I tried putting `neverwrap` in my `[options]` but it seems to have no effect. Does this work for markdown, or only asciidoc?

jnavila:
This change was specific to asciidoc. I don't think it has been ported to markdown

PeterNjeim:
It actually does work for markdown, I edited my message some minutes before you posted yours, so you probably didn't get a chance to see it.

PR #248, opened 2020-06-24
Update zh_CN.po

taotieren:


mquinson:
Thanks for this contribution, it's now integrated.

mquinson:
Btw, @taotieren , three strings were added in the meanwhile. It would be wonderful if you could update your translation, please. The easiest for me is when you use https://hosted.weblate.org/projects/po4a/ but a pull request as you did is also OK, of course.

Thanks in advance,
Mt

PR #274, opened 2020-10-15
asciidoc: process definition list continuations with indent

### LINKED ISSUE #273: asciidoc: online definition list continuations are not correctly handled
A definition such as:

```
term:: beginning of the definition
  that continues with a shift on the second line
```

I haven't found any rule about this, but some texts are using it and is seems to be handled by both python asciidoc and asciidoctor.

jnavila:
Incidentally, it seems to fix other issues with title detection.

closes #273


mquinson:
Thanks, that's just perfect.

PR #275, opened 2020-10-23
Updae simplified Chinese translator

taotieren:


mquinson:
I would be surprised if this would work as is. Please double-check https://po4a.org/man/man7/po4a.7.php.zh_CHS#lbAK

If you just want to add your content to the end of the translated file, please use `mode=eof`. Otherwise, you need to add a boundary that matches your translation content. Here, you have `position=^=head1 AUTORES`. I think that this is the spanish or catalan boundary? I'm not fluent in the language that you copy/pasted, but that's definitively not Chinese :)

Thanks, Mt

PR #277, opened 2020-10-25
Update zh_CHS.po

taotieren:


mquinson:
Hello,

was this file translated on weblate as it seems? If so, there is no need to submit it as a separate pull request: weblate merges it automatically. There is indeed recent changes to a file named zh_Hans, and I have no idea whether this is the same file or not.

Could you please tell me what I should do of these two files? I'll merge if they are different, and use the weblate version if they are the same. Am I missing something?

Thanks for your work on po4a,
Mt

taotieren:
The two files are the same, and weblate updates are slow. I will manually PR.

PR #284, opened 2020-12-20
Texinfo: add @tindex support

akngw:
Added support for `@tindex` command. It includes a test for it.

### DIFF
--- a/lib/Locale/Po4a/Texinfo.pm
+++ b/lib/Locale/Po4a/Texinfo.pm
@@ -380,7 +380,7 @@ sub translate_buffer_ignore {
 $translate_buffer_env{"ignore"} = \&translate_buffer_ignore;
 
 foreach (
-    qw(appendix section cindex findex kindex opindex pindex vindex subsection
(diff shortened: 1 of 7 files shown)


mquinson:
Thanks a lot! I'm so glad to receive a contribution to the Texinfo format, you are my hero. I've lost my understanding of this module, so it is really great. 

Are you actually using this module? I was thinking of moving it away with the other ongoing modules, the ones that don't really work yet.

Thanks again,

akngw:
Thank you for merging! I'm a po4a newbie. When I used it for a certain file,  the file seemed to be corrupted,  so I made this patch and sent it. However, the file is still corrupted...

mquinson:
I'm sorry to hear that it did not suffice for you, but I must say I'm not surprised. I'm often thinking that it'd be easier to reimplement the texinfo module from scratch, as an easy line-wise parser even if it'd break the ability to deal with advanced features of that format.

If you want to give a try to this way, I'd be glad to guide you through the internals of po4a to see how it could be done.

Issue #285, opened 2020-12-20
Build Failure due to Po4aBuilder resetting PERL5LIB.


deatondg:
On my system (Homebrew perl with custom PERL5LIB), `./Build` fails due to various include errors because my PERL5LIB variable has been reset. The issue is in Po4aBuilder.pm, and I have fixed the errors in a PR.

mquinson:
Since #286 was merged, this issue should be closed. Feel free to reopen if I'm wrong.

PR #286, opened 2020-12-20
Fix Po4aBuilder to use -I instead of reseting PERL5LIB

deatondg:
Fixes issue #285.


mquinson:
That's integrated, thanks a lot.

PR #292, opened 2021-02-08
Make the Yaml module follow the reference style of YAML Front Matter in Markdown module

### LINKED ISSUE #289: po4a command in split mode failed to extract messages that belong to an yaml master file
This is my config file _po4a.cfg_:
```
[po4a_paths] po/$master.pot $lang:po/$lang/$master.po
[type: text] content/_index.md $lang:content/$lang/_index.md master:file=messages opt:"-o markdown"
[type: yaml] i18n/en.yaml $lang:i18n/$lang.yaml master:file=messages-shared 
[options] -M utf-8 -L utf-8
```
I run it with `po4a po4a.cfg` command. The output is:
```
 (204 entries)
msggrep did not create '/tmp/po4aNSUH.pot'. Used command:
  msggrep -N "i18n/en.yaml" -o /tmp/po4aNSUH.pot /tmp/po4a-temp-pot-mYvh.pot
Please report that bug, along the information allowing to reproduce it.
```
I'm using v0.61 but as I can see the behavior in v0.62 is the same: for yaml files, message references are key names, not file names and line numbers. My guess is that since there's no information about source files, `msggrep` cannot do what is expected.

PhuNH:
- Fix #289
Instead of key names, use the file name and line number 0 as the reference. Put key names in the comment for type
- Add a test for the case of split settings with an YAML master doc

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -11,6 +11,10 @@ __   __/ _ \ / /_|___ /
 Texinfo:
  * Add support for @tindex (Github's #284)
 
+Yaml:
+ * Follow the reference style of YAML Front Matter in Markdown module to
+   fix the GitHub issue #289. (GitHub's #292)
+
 Build scripts:
  * Fix Po4aBuilder to use -I instead of reseting PERL5LIB (Github's #286)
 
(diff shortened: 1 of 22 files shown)


mquinson:
This is a very good change, many thanks for that. And sorry for not being so reactive these days, life is quite packed.

Thanks again for your contribution.

PR #296, opened 2021-02-28
asciidoc: detect and process description sublist

jnavila:


jnavila:
Sorry, I hadn't pushed everything.

Issue #136, opened 2018-06-29
Several lists of modules are not consistent


mquinson:
I have the feeling that the list of modules are not consistent nor accurate. I'd really appreciate if someone could double check that. We have one on the website, one in the README.md and one in the doc somewhere (can't remember where exactly).

Also, one should check whether these lists are accurate wrt to the actual list of modules that we have under lib/Locale/Po4a. Some modules (such as the text module) serve as an umbrella to several sub-types (such as fortune for the text module). It'd be excellent if the lists could show that too.

Thanks in advance,

mquinson:
I see:
 - the webpage: https://github.com/mquinson/po4a-website/blob/master/src/index.php.en#L42
 - the README: https://github.com/mquinson/po4a/blob/master/README.md
 - the po4a(7) manpage: https://github.com/mquinson/po4a/blob/master/doc/po4a.7.pod#supported-formats

Unifying their format would indeed help to maintain them, and that should be done :)
I see this issue as a good entry point for newcomers, and I think that seasoned developers should leave it alone, please.

bexelbie:
Thanks for catching the third one.  I agree that this is a great beginner task.  It is Hacktober after all :)

mquinson:
These list should also be compared to the list of installed modules in the tarball:
 - [ ] https://github.com/mquinson/po4a/blob/master/MANIFEST#L17

PR #312, opened 2021-07-13
Update documentation for module list

VietThan:
Also delete erroneous comment in KernelHelp module

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,9 +14,11 @@ module for the following formats:
   - dia: uncompressed Dia diagrams.
   - docbook: DocBook XML.
   - guide: Gentoo Linux's XML documentation format.
+  - halibut: Simon Tatham's documentation production system.
   - ini: INI format.
   - kernelhelp: Help messages of each kernel compilation option.
   - latex: LaTeX format.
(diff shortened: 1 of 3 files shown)


mquinson:
Hello, many thanks for this contribution! That's a long awaited one, related to #136.

If I'm correct, you did not change all the lists that were pointed in #136. Is it that they were already uptodate, or are they still to be considered?

VietThan:
Thanks @mquinson , I've changed all the ones in this repo. And yeah, I referenced the issue number with the pull request commit so GitHub will make that link automatically :D.
I'll make a separate pull for the one in the [po4-website](https://github.com/mquinson/po4a-website/blob/master/src/index.php.en#L42) repo soon. I don't know but I don't think there is a way to do one commit, one pull request for two repos

mquinson:
Indeed, it's impossible to have one PR applied to both repository, but for sake of simplicity, all bugs are counted on the po4a repo, even the ones applying to po4a-website. So I'm leaving #136 open until it gets fixed in po4a-website too.

Thanks for your time,

Issue #308, opened 2021-06-28
AsciiDoc: comments are ignored, breaking markup


paususe:
Let's say you have something like this valid AsciiDoc in source.adoc:

```
some text in English
+
// COMMENT THAT WILL CONFUSE PO4A
+
more text in English
```

This is rendered by AsciiDoc as:

```
some text in English

more text in English
```

After AsciiDoc => gettext => AsciiDoc conversion, this is what you get:

```
some text in English
+
+
more text in English
```

Now there are two consecutive "+" signs, which AsciiDoc understands as "the first + sign is to join those lines 1 and 3 in the same paragraph, and I am asked to print the second + sign", so it renders like this:

```
some text in English
+ more text in English
```

I am attaching a full example.

[po4acomments.zip](https://github.com/mquinson/po4a/files/6724221/po4acomments.zip)

jnavila:
It seems that comments management of Po4a does not follow the asciidoc "norm" according to asciidoctor. In the documentation of asciidoctor, it [is written](https://docs.asciidoctor.org/asciidoc/latest/comments/#comment-lines) that single line comments break the block structures, where as in Po4a tests, [a line comment does not split a paragraph](https://github.com/mquinson/po4a/blame/master/t/fmt/asciidoc/Comments.adoc#L14-19).

jnavila:
In fact, I'm not even sure that the behavior is consistent between asciidoc and asciidoctor here.

mquinson:
Hello all. @jnavila just commited a potential fix for this issue.

@paususe, could you please confirm that this problem is fixed by the current git version? 

Thx,

mquinson:
@paususe do you think that we could close this bug, or is there any remaining issue?
@jnavila what's your feeling here?

Thanks to both of you,
Mt

jnavila:
The changeset passed the modified unittests. But there may be some corner cases in the real world that were missed.

mquinson:
If so, I think we can close this bug. @paususe feel free to reopen it on need, or to submit another one.

Thanks a lot,

PR #320, opened 2021-09-14
asciidoc: pass through comments to translated content

### LINKED ISSUE #307: Asciidoc: Handling of comments
I'm translating Weechat documentation using po4a. Source files are in .adoc format and get processed with Asciidoctor to produce HTML.

Asciidoctor is using some tags in comments like this:

```
// tag::standard[]
   some multiline asciidoc text
// end::standard[]

// tag::debug[]
   some multiline asciidoc text
// end::debug[]
```

After po4a processing, these tags get into the comments of appropriate .po msgid strings, but discarded on generation to adoc.

Without these tags, Asciidoctor is producing errors.

Is there a way to simply pass the comments to the output, at the same position as where they appear in the original document?

jnavila:
Comments must appear translated contents to preserve document
structure.

fix #307 and #308


mquinson:
This change set is very impressive, wow. The diff is very short, but its impact is really great.

Many thanks for that.

PR #331, opened 2021-11-24
[Yaml] add option to parse only specific Yaml paths

oliverrahner:
I have the need to filter which specific paths should be extracted from my Yaml files, not only on the key name.

To stay backwards compatible I didn't modify the handling of the `keys` parameter, but instead introduced a new one.

Example:

`/usr/src/po4a/po4a-gettextize -d -f yaml -m en/antora-playbook.yml -o paths="site title,site url,asciidoc attributes idseparator" -o keys="url" -o skip_array`

This would include **all** keys named `url` and additionally the specific elements addressed by the paths:

```yaml
site:
  # matched by paths
  title: User Manual
  # matched by both
  url: https://manual.example.com/en
  start_page: index.adoc
content:
  sources:
  # matched by keys
   - url: ..
      branches: HEAD
      start_path: en/
ui:
  bundle:
    # matched by keys
    url: https://files.example.com/ui-bundle-release.zip
    snapshot: true
asciidoc:
  attributes:
    idprefix: ""
    # matched by paths
    idseparator: "-"
    kroki-fetch-diagram: true
    kroki-plantuml-include: style.puml
  extensions:
    - asciidoctor-kroki
```

### DIFF
--- a/lib/Locale/Po4a/Yaml.pm
+++ b/lib/Locale/Po4a/Yaml.pm
@@ -89,10 +89,14 @@ sub walk_yaml {
             if ( ref $el->{$key} ne ref "" ) {
                 &walk_yaml( $self, $el->{$key}, "$ctx $key" );
             } else {
-                print STDERR "working on path '$ctx $key'\n" if $self->{'options'}{'debug'};
                 my $path = "$ctx $key" =~ s/^\s+|\s+$//gr;
-                next if (  !(( $self->{options}{keys}  eq "" ) or ( exists $self->{keys}{ lc($key) }   )) or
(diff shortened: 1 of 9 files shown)


mquinson:
Hello there,

thanks for this pull request, that's really welcomed. But unfortunately, I cannot merge it until we have a sufficient set of tests covering this new feature. The CONTRIBUTING file shall explain how to add a new test, if I remember correctly.

Thanks again for your interest in po4a and for your contribution,
Mt

oliverrahner:
I added two tests, which cover the two new cases
- no keys options, no path option set (already existed)
- only keys option set (already existed)
- keys **and** path options set
- only path option set

mquinson:
Merged, thanks.

Issue #343, opened 2021-12-22
Incorrect handling of asciidoc tables with empty elements


mquinson:
Hello,

I've got a bug report from Debian which reads as follows. Please see https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1002303 for the original bug report, which I inline below.

[Petter Reinholdtsen] 
I ran into something while setting up po4a in the LinuxCNC project,
which I suspect is a bug in po4a handling of AsciiDoc tables.

This set of test files demonstrate the problem, where the translated
table format is messed up (lacking newlines).  Note, adding '.' at the
end of the lines with directions avoid the problem.

```
% cat table.adoc
= table test

.HAL Option Names
[width="100%", cols="<3s,4*<"]
|===========================================================
|Pin and Parameter Types: |HAL_BIT |HAL_FLOAT |HAL_S32 |HAL_U32
|Pin Directions:          |HAL_IN  |HAL_OUT   |HAL_IO  |
|Parameter Directions:    |HAL_RO  |HAL_RW    |        |
|===========================================================
% cat po4a.cfg
[po4a_langs] fr
[po4a_paths] table.pot $lang:$lang.po

[po4a_alias:AsciiDoc_def] AsciiDoc opt:"--keep 0 --option 'entry=lang' --option 'tablecells'"

[type: AsciiDoc_def] table.adoc $lang:$lang/table.adoc
% po4a po4a.cfg; cat fr/table.adoc 
= table test

.HAL Option Names
[width="100%", cols="<3s,4*<"]
|===========================================================
|Pin and Parameter Types: |HAL_BIT |HAL_FLOAT |HAL_S32 |HAL_U32
|Pin Directions:          |HAL_IN  |HAL_OUT   |HAL_IO  ||Parameter Directions:    |HAL_RO  |HAL_RW    |        ||===========================================================
%
```
Note, a workaround for the issue is to place a space character after the
last '|' character on the table rows with empty fields in the right
column.

jnavila:
Oh, yes... Empty cells are quite unexpected.

mquinson:
Follow up from Petter on the Debian bug:

I came across another problematic table.  Seem to me the AsciiDoc parser
interpret zero as empty in the table cells.  This table:
```
.Look Up Table
[width="50%",cols="6*^",options="header"]
|====================================
|Bit 4|Bit 3|Bit 2|Bit 1|Bit 0|Output
|0|0|0|0|0|0
|0|0|0|0|1|1
[...]

PR #345, opened 2021-12-23
asciidoc: follow up on #343, disable tablecells for non-PSV tables

jnavila:
In such case, the table is processed as a simple verbatim text block.

Please note that I added a warning message, but running `./Build binpo` keeps telling me that `po/bin/po4a.pot` is uptodate.


mquinson:
Great, thanks!

jnavila:
Can you check why po4a.pot does not update?

mquinson:
I'm not sure. I usually do `./Build postats` to update the POT files, and then I commit the POT files and revert the PO ones. I prefer to have weblate doing the msgmerge, even if I forgot what could possibly go wrong if I do it manually.

I think that this is not a big deal, as I don't mind updating the POT files on need. I feel too lazy to go and fix the generation scripts instead... I hope you don't mind

jnavila:
Oh, ok. `postats` then.

PR #355, opened 2022-01-23
asciidoc: Support Hugo shortcode on macros

### LINKED ISSUE #352: Asciidoctor/Include issue, when using Hugo shortcode
[Source file](https://raw.githubusercontent.com/dbaio/po4a-debug/main/po4a-asciidoctor-includes.adoc).
[PO file](https://raw.githubusercontent.com/dbaio/po4a-debug/main/po4a-asciidoctor-includes.po).
po4a version 0.66 and prior releases has the same issue.

```
po4a-gettextize \
        --format asciidoc \
        --option compat=asciidoctor \
        --option yfm_keys=title,part,description \
        --master "po4a-asciidoctor-includes.adoc" \
        --master-charset "UTF-8" \
        --copyright-holder "The FreeBSD Project" \
        --package-name "FreeBSD Documentation" \
        --po "po4a-asciidoctor-includes.po"
```

-->

```
#. type: Plain text
#: po4a-asciidoctor-includes.adoc:35
msgid ""
"include::shared/attributes/attributes-{{% lang %}}.adoc[] "
"include::shared/{{% lang %}}/teams.adoc[] include::shared/{{% lang "
"%}}/mailing-lists.adoc[] include::shared/{{% lang %}}/urls.adoc[]"
msgstr ""
```

Includes without `{{% lang %}}` works fine; they are skipped.

I'm reporting this so others can help, but the plan is to dig/debug po4a later.

https://docs.asciidoctor.org/asciidoc/latest/directives/include/

dbaio:
Closes #352

### DIFF
--- a/lib/Locale/Po4a/AsciiDoc.pm
+++ b/lib/Locale/Po4a/AsciiDoc.pm
@@ -790,7 +790,7 @@ sub parse {
             }
             @comments = ();
         } elsif ( not defined $self->{verbatim}
-            and ( $line =~ m/^([\w\d][\w\d-]*)(::)(\S*)\[(.*)\]$/ ) )
+            and ( $line =~ m/^([\w\d][\w\d-]*)(::)(\S*|\S*\{.*\}\S*)\[(.*)\]$/ ) )
         {
             my $macroname   = $1;
             my $macrotype   = $2;
--- a/t/fmt-asciidoc.t
+++ b/t/fmt-asciidoc.t
@@ -12,7 +12,7 @@ use Testhelper;
 my @tests;
 
 foreach my $t (
-    qw(Titles BlockTitles BlockId Paragraphs DelimitedBlocks Lists Footnotes Callouts Comments Tables TablesImageText Attributes StyleMacro)
+    qw(Titles BlockTitles BlockId Paragraphs DelimitedBlocks Lists Footnotes Callouts Comments Tables TablesImageText Attributes StyleMacro MacroIncludesHugo)
   )
 {
     push @tests, { 'format' => 'asciidoc', 'input' => "fmt/asciidoc/$t.adoc" };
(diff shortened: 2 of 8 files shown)


t/fmt-asciidoc.t
```diff
@@ -55,6 +55,11 @@ push @tests,
     'format'  => 'asciidoc',
     'options' => "-o yfm_skip_array -o yfm_keys='title , subtitle,paragraph'",
     'input'   => "fmt/asciidoc/YamlFrontMatter_Option.adoc",
+  },
+  {
+    'format'  => 'asciidoc',
+    'input'   => 'fmt/asciidoc/MacroIncludesHugo.adoc',
+    'doc'     => 'asciidoctor includes with hugo shortcode',
   };
```
jnavila, line 63: The `MacroIncludesHugo.adoc` should simply be included in the list on line 15 as it does not require a special command line option.

lib/Locale/Po4a/AsciiDoc.pm
```diff
@@ -790,7 +790,7 @@ sub parse {
             }
             @comments = ();
         } elsif ( not defined $self->{verbatim}
-            and ( $line =~ m/^([\w\d][\w\d-]*)(::)(\S*)\[(.*)\]$/ ) )
+            and ( $line =~ m/^([\w\d][\w\d-]*)(::)(\S*|\S*\{.*\}\S*)\[(.*)\]$/ ) )
```
jnavila, line 793: This only matches one transclusion. Is there any chance that several Hugo shortcodes may be used?

jnavila (changes requested):
This change is quite benign.


Apart from the small change in the test, LGTM.

lib/Locale/Po4a/AsciiDoc.pm
dbaio: In all macro examples I found, there is always one variable.
I think it would be very unusual having more (but it's possible).

t/fmt-asciidoc.t
dbaio: Ok, I'll change that.

jnavila (approved):

Issue #299, opened 2021-03-20
Asciidoc: Better message splitting needed


mariobl:
Hello,

currently I'm working on the translation of man pages written in an Asciidoctor-compatible format of Asciidoc. Although [Locale::Po4a::AsciiDoc(3pm)](https://po4a.org/man/man3/Locale%3A%3APo4a%3A%3AAsciiDoc.3pm.php.en) says "Tested successfully on simple AsciiDoc files", it works much better than expected, however, with some constraints... I generate the *.pot file with the following command:

`po4a-gettextize -f asciidoc -o compat=asciidoctor -m nsenter.1.adoc -p nsenter.1.pot`

The source file contains the following paragraph:

```
*-a*, *--all*::
  Enter all namespaces of the target process by the default _/proc/[pid]/ns/+++*+++_ namespace paths. The default paths to the target process namespaces may be overwritten by namespace specific options (e.g., *--all --mount*=[_path_]). +
  {nbsp} +
  The user namespace will be ignored if the same as the caller's current user namespace. It prevents a caller that has dropped capabilities from regaining those capabilities via a call to setns(). See *setns*(2) for more details.
```

The `{nbsp} +` is needed to get an empty line within an indented part, here an option description. I would actually expect that Po4a splits it in its logical parts, but I get this:

```
#. type: Plain text
#: nsenter.1.adoc:54
msgid ""
"Enter all namespaces of the target process by the default "
"_/proc/[pid]/ns/+++*+++_ namespace paths. The default paths to the target "
"process namespaces may be overwritten by namespace specific options (e.g., "
"*--all --mount*=[_path_]). + {nbsp} + The user namespace will be ignored if "
"the same as the caller's current user namespace. It prevents a caller that "
"has dropped capabilities from regaining those capabilities via a call to "
"setns(). See *setns*(2) for more details."
msgstr ""
```

This is annoying, because translators could be confused by the "+ {nbsp} +" – and it becomes more difficult to reuse existing translations based on *roff sources.  After translating, I get this in the resulting *roff file:

```
[...]

PR #362, opened 2022-06-28
asciidoc : manage linebreaks

jnavila:
This change passes on a minimal test file, but not sure to cover all the cases described in   #299.

A good sanity check would be for translators using this feature to report their mileage.


mquinson:
Thanks a lot for your work, @jnavila. I think I'll manage to touch base at some point, sorry for the inconvenience.

PR #363, opened 2022-06-30
asciidoc: fix linebreaks handling

jnavila:
The correct sequence is SPACE + PLUS + CR

This should end the fixup for  #299


mquinson:
I seem to understand that the space becomes mandatory before the + at CR. Is it intended? It'd find more sensible to make it optional, don't you think?
But I may well be utterly wrong, as I didn't follow #299 closely. If so, please bear with me.

jnavila:
From [the specification](https://docs.asciidoctor.org/asciidoc/latest/blocks/hard-line-breaks/), the space is mandatory, there are no exceptions, so it's not optional. Otherwise, it could be interpreted as an inline format for `monospace` [text](https://docs.asciidoctor.org/asciidoc/latest/text/literal-monospace/).


The mistake was indeed mine in the first PR.

mquinson:
I'm merging, but maybe we should add a point to the NEWS file?

jnavila:
It's not needed. This PR is just the fixing of the previous one which was half-baked.

PR #366, opened 2022-07-09
Add new xhtml table test and stop claiming tables are not tested

petterreinholdtsen:
The Xhtml documentation indicate that table support is not present
as it is not tested.  Tables seem to work, so lets add a test
and remove the documentation comment.


mquinson:
This is a great pull request, thanks. Merged.

In the future, if you want to reach the 102% of perfection, please add a little word in the NEWS file.

Thanks again,
Mt

PR #387, opened 2022-09-25
Avoid embedding the build date in compressed manpages

raboof:
Since the manpages are generated at build time, and gzip by
default includes the file timestamp in the compressed file,
this 'leaks' the build time into the resulting artifact.

There are many advantages to avoiding 'leaking' the build time
into the artifact: it can make detecting compromised CI servers
easier through https://reproducible-builds.org/ , can make
content-addressed storage more effective and improve cache hits
when rebuilding.


mquinson:
Thanks for this change, it's now integrated.

Issue #389, opened 2022-12-08
XML: Placeholder for empty element


k-yaegashi:
Given the following XML source
```xml
<?xml version="1.0" encoding="utf-8"?>
<doc>
  <p>
    A <place attr="value1"/> B <place attr="value2"/> C <place attr="value3">and</place> D.
  </p>
</doc>
```
and this configuration,
```conf
[po4a_langs] ja
[po4a_paths] test.pot ja:ja.po

[po4a_alias:spec] xml opt:"-o \"placeholder=<place>\""

[type: spec] test.xml ja:output.xml
```
I expected something like the following message catalogs:
```po
#. type: Content of: <doc><p><place>
#: test.xml:4
#, no-wrap
msgid "and"
msgstr ""

#. type: Content of: <doc><p>
#: test.xml:4
#, no-wrap
msgid ""
"    A <placeholder type=\"place\" id=\"0\"/> B <placeholder type=\"place\" id=\"1\"/> C <placeholder type=\"place\" id=\"2\"/> D.\n"
"  "
msgstr ""
```
However, the actual result with the latest po4a 0.68 is:
```po
#. type: Content of: <doc><p>
#: test.xml:4
#, no-wrap
msgid "    A "
msgstr ""

#. type: Content of: <doc><p>
#: test.xml:4
#, no-wrap
msgid " B "
msgstr ""

#. type: Content of: <doc><p><place>
#: test.xml:4
#, no-wrap
msgid "and"
msgstr ""

#. type: Content of: <doc><p>
#: test.xml:4
#, no-wrap
msgid ""
" C <placeholder type=\"place\" id=\"0\"/> D.\n"
"  "
msgstr ""
```
It means placeholder doesn't work for empty element, that is not convenient when empty and non-empty elements are mixed.

In order to get the desirable result, the source XML must be
```xml
<?xml version="1.0" encoding="utf-8"?>
<doc>
  <p>
    A <place attr="value1"></place> B <place attr="value2"></place> C <place attr="value3">and</place> D.
  </p>
</doc>
```
But in many cases, it is nonsense to change source code only for translation.

mquinson:
Fixed by PR #390.

PR #390, opened 2022-12-09
Improved XML inline-related translation features

k-yaegashi:
Resolves the issues #389 and #391.

This fix can
- treat placeholder for empty elements, and
- translate attributes inside attribute-folded inline and placeholder.

In order to realize former feature,
- extracted special processes for open and close tags in `treat_content` and put into `treat_content_open_tag` and `treat_content_close_tag` respectively,
- applied both `treat_content_open_tag` and `treat_content_close_tag` for empty-element tag in `treat_content`, and
- disabled tag breaking for placeholder-marked empty-element tag in `tag_break_alone`.

In order to realize latter feature, when processing open tag marked as placeholder or attribute-folded, passed the tag into `treat_attributes`.

### DIFF
--- a/lib/Locale/Po4a/Xml.pm
+++ b/lib/Locale/Po4a/Xml.pm
@@ -1071,7 +1071,7 @@ sub CDATA_trans {
 sub tag_break_alone {
     my ( $self, @tag ) = @_;
     my $struct = $self->get_path( $self->get_tag_name(@tag) );
-    if ( $self->get_translate_options($struct) =~ m/i/ ) {
+    if ( $self->get_translate_options($struct) =~ m/[ip]/ ) {
         return 0;
     } else {
         return 1;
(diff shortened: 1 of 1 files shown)


mquinson:
Hello @k-yaegashi,

thanks for these changes, they seem very good. I hope to see more such changes in the future ;)

It would be easier for me if you could split your changes into several commits. Here for example, it'd be better to have one commit splitting the functions apart, and another commit changing this code to implement the new feature.

Another thing that would be really welcome the next time that you commit to this project (soon, I hope) is that any new feature must be tested accordingly. This one should be easy for you since you already have the test files in the bug reports (which are very good, thanks for that).

Again, I don't want to criticize your work. It's only that maintaining po4a is unfortunately very low priority for me right now and I desperately need help from new contributors like you to keep it going. Thanks again for this change, and I'm waiting for the next one ;)

PR #397, opened 2023-01-01
Fix linux github action

jnavila:
Signed-off-by: Jean-Noël Avila <jn.avila@free.fr>

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


mquinson:
Why is this useful? Could you point me to the relevant doc? Thx,

jnavila:
Until now, `checkout@v2` could be used to check out the project in the Docker CI, but it's based on node 12, which has been deprecated. Now, we have to use `checkout@v3`, which makes the "Build on Linux CI" check work. Otherwise, we no longer have any CI checkout our back.

Checkmarx Scan is still failing, but I don't really know what it is, whether it ever worked or if it is still useful.

mquinson:
Thanks, merged. Checkmarx was just an experiment, I removed it in between.

Issue #400, opened 2023-03-26
AsciiDoc exclude cross-references and anchors ids from the *.po(t) translation file


benoitrolland:
readinf the [Locale::Po4a::AsciiDoc](https://po4a.org/man/man3/Locale::Po4a::AsciiDoc.3pm.php) documentation,
I do not see how to exclude anchors ids   and cross reference ids from the generated translation file.
I believe this prevent an effective automatic translation of asciidoc files using po4a.
LEt find the details [here](https://stackoverflow.com/questions/75848327/po4a-asciidoc-exclude-from-translation)

mquinson:
Hello @benoitrolland, thanks for the feedback. 

Could you please provide me with an example file exemplifying what you want to exclude? I'm not expert in all the formats handled by po4a and often struggle to generate such test cases myself.

Thanks,

benoitrolland:
> Hello @benoitrolland, thanks for the feedback.
> 
> Could you please provide me with an example file exemplifying what you want to exclude? I'm not expert in all the formats handled by po4a and often struggle to generate such test cases myself.
> 
> Thanks,

from an asciidoc file containing:

```
[[ill-sketches-intro,Sketches]]
[NOTE,icon=texte-introduction.svg]
.Sketches
====
. <<img-Things-various>>
====
```

(...)

```
[[img-Things-various, Various things]]
.<<ill-sketches-intro,Sketches: >><<img-Things-various>>
[caption=""]
image::intro/07ThingsVarious.jpg[img-Things-various,180,100,float="left",align="center"]
```
Using po4a version 0.69.,

the generated *.pot file content is:

```
    #. type: Block title
    #: bookname.adoc.pp:343
    #, no-wrap
    msgid "<<ill-sketches-intro,Sketches: >><<img-Things-various>>"
    msgstr ""
    
    #. type: Positional ($1) AttributeList argument for macro 'image'
    #: bookname.adoc.pp:345
    #, no-wrap
    msgid "img-Things-various"
    msgstr ""
    
    #. type: Target for macro image
    #: bookname.adoc.pp:345
    #, no-wrap
    msgid "intro/07ThingsVarious.jpg"
    msgstr ""

    #. type: Block title
    #: bookname.adoc.pp:2469
    #: bookname.adoc.pp:2483
    #: bookname.adoc.pp:2492
    #: bookname.adoc.pp:2517
    #: bookname.adoc.pp:2545
    #: bookname.adoc.pp:2576
    #, no-wrap
[...]

PR #402, opened 2023-04-02
asciidoc: process xlabel of id

### LINKED ISSUE #401: Locale::Po4a::AsciiDoc is pot_in: supported
First, thank you so much for the great job of including assciidoc into po4, allowing translation from asciidoc files.

Willing to exclude some elements from translation, I am trying to use the _pot_in feature_ as described in the [doc](https://po4a.org/man/man1/po4a.1.php#lbAN
)
When specifying the path to a copy of the originaly generated pot file as value to pot_in: in the po4a.cfg file, 
a new pot file is generated on next run and is different from the first.  
More of that it seems corrupted and as such is not usable within  Poedit, 
since the elements to be translated are full of # followed by what should be the comments  .
msgid "" seems duplicated and its first occurence included within the normally preceeding comments:

#. type: Plain text
#: po4a/book-I.adoc.pp.filtered.pot:149
#: po4a/book-II.adoc.pp.filtered.pot:139
#: po4a/book-III.adoc.pp.filtered.pot:139
msgid ""
"#. type: Plain text #: book-I.adoc.pp:174 Book-"
"II.adoc.pp:164 #: book-III.adoc.pp:164 #, no-"
"wrap msgid \\"\\" \\"This is a text to not show the real one"
"but technically this is, \\\\\\"the_same_\\\\\\", I hope you"
"do not mind."

jnavila:
This partly fixes #401 but not completely. The problem is that po format does not support inline markup, so we cannot hide formatting marks away from translators. We could use a different format for the translation (for instance xml), but this would require a lot of changes in AsciiDoc.pm.


benoitrolland:
> This partly fixes #401 but not completely. The problem is that po format does not support inline markup, so we cannot hide formatting marks away from translators. We could use a different format for the translation (for instance xml), but this would require a lot of changes in AsciiDoc.pm.

Looking to the changes I guess it more likely fixes #400 than #401

mquinson:
Thanks to both of you for this fix, it's now integrated.

PR #406, opened 2023-05-01
TeX.pm: allow \input command surrounded by curly braces

### LINKED ISSUE #405: latex: `\input` breaks when surrounded with {curly braces}
Minimal reproducer (`main.tex`, with at least an empty `subfile.tex` next to it):

```
\documentclass{article} 
 
\begin{document} 
{\input{subfile.tex}}
\end{document}
```

See that `\input` is surrounded by curly braces. Compilation fails when that is the case, and succeeds otherwise.
As to why I need this: I use multiple `tikzpicture` graphs in `subcaptionbox` that I input from other files.

This is the (confusing) error:

```
Use of uninitialized value $newfilepath in string eq at /usr/share/perl5/vendor_perl/Locale/Po4a/TeX.pm line 998.
po4a::tex: Cannot find main.tex with kpsewhich
```

Even more confusingly, the exact same error is displayed when `subfile.tex` does not exist (with or without curly braces), and complains that the parent file cannot be found, while it's the included file that does not exist.

MayeulC:
Previously, the closing brace `}` would be matched as part of the filename, which was then not found.

This allows the use of `\input` commands as macro parameters, for instance with the `\subcaptionbox` command:
```tex
\subcaptionbox{Legend}[2cm]{\input{sometikzplot.tex}}
```
Fixes #405

### DIFF
--- a/lib/Locale/Po4a/TeX.pm
+++ b/lib/Locale/Po4a/TeX.pm
@@ -969,7 +969,7 @@ sub read_file {
         while (
             $textline =~ /^((?:[^%]|(?<!\\)(?:\\\\)*\\%)*)
                               \\(include|input)
-                              \{([^\{]*)\}(.*)$/x
+                              \{([^\{\}]*)\}(.*)$/x
           )
         {
             my ( $begin, $newfilename, $end ) = ( $1, $3, $4 );


mquinson:
Hello @MayeulC, many thanks for reporting this issue and providing the fix. Could you please also add a little test for this feature? You could either extend an existing test, or add a new one. Please see https://github.com/mquinson/po4a/blob/master/CONTRIBUTING.md#writing-a-test

MayeulC:
Hi, I'll try to, I also figured it would be nice to provide a test, thank you for directing me towards the relevant documentation.

I don't have a clear timeline: I may have a look tomorrow or later this week.

mquinson:
To be perfectly honest, I hope that the doc is still uptodate. That's a very long time that I didn't review it. Do not hesitate to ask if you have any problem.

Beside of this, this is a contribution to an open source project. There is absolutely no deadline. Your contribution will be really welcome, no matter when it occurs.

Thanks!

mquinson:
Hello @MayeulC, if you could provide me the LaTeX file needed for the test, I'll integrate it in the test infrastructure. That'd be a live example of how to do that if the doc is outdated.

The difficult part for me is to come up with the LaTeX file that correctly tests the feature.

mquinson:
@lksj while you're at the TeX module, could you please take care of this PR? I think it's only missing a test or two.

Thanks in advance if you manage to do something...

PR #407, opened 2023-05-01
TeX.pm: allow verbatim environment declaration without trailing space

MayeulC:
Previously, a trailing space was matched by \s+, but this is counter-intuitive, and couldn't make it work before digging in the code.

Match the behavior of non-verbatim environment by making it \s* (allowing, but not requiring trailing spaces).


mquinson:
Thanks a lot, I'll merge this one as soon as the tests are done.

PR #415, opened 2023-05-19
asciidoc: fix processing of multiline attributes

### LINKED ISSUE #414: Asciidoc / Only first line of a multiline attributes is recognized as assignment
When I define a multiline attribute in an adoc file like this:

```
:myvar: {var1} + \
{var2} + \
{var3}
```
I got an po-file output like this:

```
#. type: Plain text
#: accessories.adoc:13
msgid "{var2} + \\ {var3}"
msgstr ""
```

At first, I thought, it is an issue with the recognition of "+ \\". After further tests I am of the opinion that the attribute is not recognized as a multiline definition. If the definition were recognized as multiline, then the output 
```{var2} + \\ {var3} ```
should not be in the po-file.

[bug.zip](https://github.com/mquinson/po4a/files/11516065/bug.zip)

jnavila:
The processing of multiline attributes was limited to the legacy syntax, and was broken for the new syntax. This commit fixes it.

Fixes #414


mquinson:
And thanks for this one too, Jean-Nöel.

PR #422, opened 2023-05-31
markdown: support link reference

gemmaro:
This pull request marks link reference definitions in the markdown document as no-wrap, which avoids unintentional line breaks at the colon in the long link reference definition line.

Here are some limitations below, but I suspect it works in most cases:

* Multiline link reference definition is not supported ([Example 193](https://spec.commonmark.org/0.30/#example-193))
* It misinterprets a link reference definition if its link text is invalid ([Example 197](https://spec.commonmark.org/0.30/#example-197))
* Link reference definition may interrupt normal paragraph ([Example 213](https://spec.commonmark.org/0.30/#example-213))

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -21,6 +21,9 @@ Tex:
  * Allow verbatim environment declaration without trailing space
    (GitHub's #407). Thanks Mayeul Cantan for the report & the fix.
 
+Markdown:
+ * Treat link reference definitions as no-wrap (GitHub's #422) [gemmaro].
+
 Translations:
  * Updated: French, thanks Jérémie Tarot.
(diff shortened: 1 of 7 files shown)


mquinson:
Hello @gemmaro, thanks for this PR. I see that it's marked as a draft, so I'll refrain from merging it, but it sounds pretty good to me. I'd love a little note in the NEWS file also, but I can do that myself afterward if you prefer.

What do you plan to do about it before unmarking it 'draft'?

Thanks again,

gemmaro:
I updated regexp pattern, to accept indentation for example, according to the CommonMark specification.
I also applied tidyall to format the code, and added about the change to the NEWS file.

Issue #416, opened 2023-05-22
Asciidoc / Keeping original formating


suddenfall:
I have feature request for the asciidoc module, that may overlap with:
https://github.com/mquinson/po4a/issues/291

I use Asciidoctor and for me it is also a problem that the content from my master documents are wrapped in the po-files and thus reformatted. I propose to introduce an option in the Asciidoc module so that the original formatting in the po files is preserved.

### Motivation
Currently, I use deepl API. Reformatting the content makes it harder to use the API optimally. In addition, reformatting makes it necessary for me to check the translated documents after rendering to PDF whether the original formatting from the master documents has been retained.

I have structured my master documents according to the principle: One line contains exactly one sentence.

That would fit perfectly to deepl API, where a text line is interpret a sentence. Unfortunately, I have not found a way with po4a or the Asciidoc module to prevent the reformatting of the content.

Therefore I have to enable in the deepl API that sentences are recognized by punctuation marks. This works, but leads to problems with index terms, for example. If I write this content:

```
== Maximus Primus
(((Primus Magnificus)))
Lorem ipsum dolor sit amet, consectetur adipisici elit, sed eiusmod tempor incidunt ut labore et dolore magna aliqua. 
Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquid ex ea commodi consequat. 
Quis aute iure reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. 
Excepteur sint obcaecat cupiditat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.
 ```

It results to following message id:

```
"(((Primus Magnificus)))  Lorem ipsum dolor sit amet, consectetur adipisici "
"elit, sed eiusmod tempor incidunt ut labore et dolore magna aliqua.  Ut enim "
"ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquid "
"ex ea commodi consequat.  Quis aute iure reprehenderit in voluptate velit "
"esse cillum dolore eu fugiat nulla pariatur.  Excepteur sint obcaecat "
"cupiditat non proident, sunt in culpa qui officia deserunt mollit anim id "
"est laborum."
```

The index term `(((Primus Magnificus)))` is prepend to the content in the same line. This can lead to upper and lower case being mixed up in the translation and even to the entire sentence being distorted in terms of sense.

### Plea for the feature
[...]

PR #424, opened 2023-06-07
asciidoc: process transparent index entries in their own segment

jnavila:
This patch addresses the issue of keeping transparent index entries in paragraphs. It does not address "visible" index entries, which are formatted with ((...)).

This addresses GH #416.


jnavila:
Well, this was a bit too fast a merge. I should have marked this as a draft, because some checks were expected from @suddenfall

mquinson:
woops, sorry.

mquinson:
I reviewed your changes, and it seemed very reasonable. I just added a comment after merging your work. If @suddenfall finds further difficulties, we can always do an additional PR. Next time, I'll be more cautious when you ask for reviews in the comments...

PR #432, opened 2023-08-04
Use xmlcatalog to find DocBook XSL file

### LINKED ISSUE #431: proposal: use xmlcatalog to find docbook.xsl
version: 0.69

use xmlcatalog(1) from libxml2 to find the xsl file rather than checking a hardcoded absolute path. it can be found with the following command and it will print the path on success

`xmlcatalog --noout "" "http://docbook.sourceforge.net/release/xsl/current/manpages/docbook.xsl"`

gemmaro:
Fix #431.

### DIFF
--- a/.tidyallrc
+++ b/.tidyallrc
@@ -2,4 +2,5 @@
 select = po4a{,-gettextize,-normalize,-translate,-updatepo}
 select = lib/**/*.pm
 select = t/*.{t,pm}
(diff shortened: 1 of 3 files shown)


mquinson:
Thanks a lot for this change, and sorry for the delay, I'm really overwhelmed these days.

Please keep up the good work ;)
Mt

PR #436, opened 2023-08-14
asciidoc: correctly handle line in paragraphs that start with a dot

jnavila:
When the asciidoc is poorly formatted, it can happen that a paragraph contains a line that starts with a dot. In this case, the line was considered as a title of the following section, and the paragraph was split in three.

This is obviously not what we want. Asciidoctor seems to cope with this, so do the same.


mquinson:
Thanks a lot, merged!

PR #451, opened 2023-12-31
Fix Chinese lang codes

stokito:
zh_CHS to zh_Hans
zh_CHT to zh_Hant

Fix #419

This caused an error in debiman so came here to fix it.

PR #452, opened 2024-01-01
Fix Chinese lang codes: remove country specific codes zh_CN and zh_HK

stokito:
After the #451 I remember to check if the zh_CN is used and indeed, some files with the incorrect lang code are exists.


mquinson:
Thanks a lot!

PR #456, opened 2024-01-12
po4a: Some small typo fixes in documentation

Fat-Zer:
probably those should be merged after the release to not mess up translations.


mquinson:
Thanks. I'll indeed merge it after the release to not complicate the translation work. :pray:

mquinson:
Thanks again!

PR #459, opened 2024-01-13
Fix Japanese language code from jp to ja

gemmaro:
I think that this is trivial and should be done after the release as it will cause the PO file to be updated.

- [ ] Update PO files


mquinson:
Thanks!

Issue #464, opened 2024-01-30
AsciiDoc: Sentences joined by a double space in a para - sentence per line


git-pear:
Hello,

I have noticed that if an AsciiDoc text paragraph is styled as 'sentence per line', the resulting .po(t) file contains double spaces between the sentences of the paragraph.

If the paragraph in asciidoc is not styled as 'sentence per line' (all the para's sentences are on one line), the .po file is "normal", without double spaces.

Is such conversion of the 'sentence per line' paragraph of asciidoc' into 'double spaced' .pot/.po intentional, with a reason behind that?

Thank you very much for looking into this matter.

Josef Hruska

git-pear:
This can be used as a sample asciidoc text:

= Test of a sentence per line para

== Para not styled as 'sentence per line'

Not sentence per line para. This para is written such as all it's sentences are placed on one line only. After po4a processing, there is just single space between the senteces of the para.

== Para styled as 'sentence per line'

Sentence per line para.
This para is written such as all it's sentences are placed each one on its own line.
After po4a processing, there are double spaces between the sentences of the para.

jnavila:
What I found is that, this is mostly due to the fact that the carriage return is replaced with a space, and you can have a dangling space at the end of the line.

I can make `po4a` tidy the input string when it is not "wrap", but that logic will upset a lot of existing translations.

git-pear:
> I can make po4a tidy the input string when it is not "wrap", but that logic will upset a lot of existing translations.

Thank you for looking into this. I consider that not worth to make others upset.

jnavila:
Maybe I can add an option, so that the default behavior is retained but you can still get a cleaned up po file. Let's try this.

git-pear:
Then thank you very much indeed. I did not think about this 'configuration' possibility.

git-pear:
> Still somewhat unsure here

Ok, this is not urgent IMHO, take your time to think it all through.

I was puzzled by the double spaces mainly because they are, let's say, highlighted in [weblate](https://weblate.org) editor I use for translating documentation. Not experienced translators may tend towards transferring the double spaces also to their translation(s), which is actually not necessary.

jnavila:
> I was puzzled by the double spaces mainly because they are, let's say, highlighted in [weblate](https://weblate.org) editor I use for translating documentation.

[...]

PR #481, opened 2024-02-28
asciidoc: add cleanspaces option to remove extra spaces

jnavila:
Extra spaces are removed in no-wrap segments, so that the output conforms to English typography rules.

This fixes GitHub #464


mquinson:
Hello Jean-Noël, thanks for this patch.

maybe the cleanup should be the default behavior for new users, don't you think?

I'm not completely decided, but it seems somehow better.

Mt

jnavila:
I don't grasp the idea of "the default behavior for new users". Do you mean that we should somehow differentiate between creating the pot file and updating it? How can we do it, and save the actual state of the option?

For sure, for existing translation project, it is critical to retain the behavior of unset cleanspaces, unless the project manager actively enables it.

jnavila:
OK, so your option is the harsh one :laughing: ... Being the main developer of this feature and being plagued with this is issue, I know that this can be cumbersome to change a lot of segments, only for some whitespace issues.

I don't know how to intend to warn user about this compatibility breakage. I would at least issue a warning when such issue is detected and extra whitespaces are detected and removed. So maybe, this option should be activated by default, generating a warning. When the option is set to true or false, the warning would disappear.

Personally, I understand your point: a sane behavior for new projects, but I don't like at all breaking the increasing number of projects relying on po4a. In any case, it would be a breaking change that needs to be strongly advertised.

jnavila:
Please note that right now, the patch cleans up the original string of its extra spaces but does not post process the translated string.

mquinson:
Thanks a lot @jnavila, I merged this as is. You are right about not upsetting the translators by default.

PR #466, opened 2024-02-07
New module for Gemtext format

gemmaro:
Hello,

This is a new module for the Gemtext format.
This plain text format is mainly used by the [Gemini protocol][gemini].
I have used this module to translate some of their documents and it seems to work.

Thank you,
gemmaro.

[gemini]: https://geminiprotocol.net/

PR #485, opened 2024-03-02
New module for Org format

gemmaro:
Hello,

This is a new module for the Org format, which is typically used in [the Org mode][org].
I have used this module to translate some of their documents (such as the [Org Mode Compact Guide][guide]), and it seems to work.

~~This also adds some links to the Gemtext module (#466).~~ I'll make another pull request.

Thank you,
gemmaro.

- - -

TODO:

- [x] I have added List::Util to the dependencies, but considering whether it is really necessary.
      It is useful for manipulating lists, but as more dependencies are added, the packaging will need to be updated.
- [x] Fixes translate() and pushline() usage.
      There are places where $ref and $type are missed or extra passed.
- [x] Regular expressions improvements, e.g.  `\A` should be used instead of `^` in some places.
- [x] Conflict resolution.

These tasks are future work:

- Conform to [the Org Syntax](https://orgmode.org/worg/org-syntax.html)
- Improved parser when targeting the Org reference manual.

[org]: https://orgmode.org/
[guide]: https://orgmode.org/guide/

### DIFF
--- a/lib/Locale/Po4a/Org.pm
+++ b/lib/Locale/Po4a/Org.pm
@@ -1,0 +1,467 @@
+#!/usr/bin/env perl -w
+
+# Po4a::Org.pm
+#
+# extract and translate translatable strings from a Org documents
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+#
+########################################################################
+
+package Locale::Po4a::Org;
+
+use 5.006;
+use strict;
+use warnings;
+
+use parent qw(Locale::Po4a::TransTractor);
+
+sub initialize {
+    my ( $self, %options ) = @_;
+
+    $self->{options}{skip_keywords}   = [];
+    $self->{options}{skip_properties} = [];
+    $self->{options}{skip_heading}    = 0;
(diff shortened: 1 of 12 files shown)


lib/Locale/Po4a/Org.pm
```diff
@@ -0,0 +1,480 @@
...
+documentation in the Org format, used by the L<Org
+Mode|https://orgmode.org/>.
+
+=head1 CONFIGURATION
+
+=over
+
+=item B<skip_keywords>
+
+Space-separated list of keywords.  They are not translated.
```
Fat-Zer, line 437:
```suggestion
Space-separated list of keywords which won't be translated.
```

lib/Locale/Po4a/Org.pm
```diff
@@ -0,0 +1,480 @@
...
+
+=item B<skip_properties>
+
+Space-separated list of properties.  They are not translated.
```
Fat-Zer, line 441:
```suggestion
Space-separated list of properties which won't be translated.
```

Fat-Zer (commented):
Just a couple of small suggestions on doc wording.

Besides that, you also probably would want to add `Org.pm` and `Gemtext.pm` to `po/pod.conf` so they would be translated.

gemmaro:
Thank you for your review.
I've applied your suggestions and added new modules to the translation targets (and updated the PO files with `./Build docpo`.)

gemmaro:
I've noticed that there are a few things that need fixing, so I'm switching to draft. Once these are fixed, I'll remove the draft status.

mquinson:
Hello @gemmaro, there is no rush, and I'm thankful for your work. 

If you feel so, please tell us what remains to be done here, so that maybe someone helps you implementing it.

gemmaro:
Hello @mquinson, thank you for checking.
I added a TODO list to my initial comment about what remains to be completed.

It will take some time to work on this for the time being.
I'll resume when the time comes, but anyone else is welcome to take over.

gemmaro:
Changes from last time are:

- Add the skip_headings option.
- Drop unrelated changes.
- Drop PO update commit.
- Cleaned up the code a bit.

I thought I should mark this as complete after checking that it works for the Org reference manual (and translate it) at the first time.  However I found that it takes too long to confirm that it supports enough syntax patterns.  The current implementation doesn't support all of them, but is hopefully sufficient for simpler documents.

Issue #474, opened 2024-02-19
How to generate Markdown without line wrap?


rffontenelle:
I have a Markdown file as master doc that has long lines. I would like to generate the translated doc with these long lines, instead of having them wrapped in +/- 75 characters per line.

Adding the `-o neverwrap` option causes the PO file to have all strings with `no-wrap` flag and with a trailing new-line `\n` character. I'd prefer to not have the translation strings without the `no-wrap` flag and without the trailing new-line character.

Is there an alternative?

Fat-Zer:
As far as I can see, as for now there is no way to do it.

It seems, it was supposed to be controlled by `po4a-translate`'s `--width` option, but it was never actually implemented.

jnavila:
This was raised (#242) and implemented (0bf9652) in Asciidoc, by removing the carriage return after the fact... This request is understandable, as, just like Asciidoc, Markdown may encounter mis-formatting issues due to unlucky line-wrapping.

PR #492, opened 2024-03-25
Add an option -w/--width to rewrap text on an arbitrary column

Fat-Zer:
The option was previously added for po4a-translate, but for more than a
decade weren't actually implemented.

This was added to resolve https://github.com/mquinson/po4a/issues/474.


mquinson:
This PR is very good, thanks @Fat-Zer (and sorry for the delay).

PR #503, opened 2024-06-04
Rename a case sensitive test fixture file

gemmaro:
Hello,

This renames a test fixture since there is a file with the same name (in case insensitive way).
This might cause a problem in a case insensitive filesystem, e.g. `git clone` fails for the file path collision.

I checked with the following script to find if there are others (and result in a single pair.)

```ruby
#!/usr/bin/env ruby
files = Set.new
Dir['**/*'].each do |file|
  File.file?(file) or next
  file.downcase!
  files.member?(file) and raise file
  files << file
end
```

Thank you,
gemmaro.


mquinson:
Thanks for this!

PR #518, opened 2024-09-03
Fix some typos

gemmaro:


mquinson:
Woops, sorry for making so many typos in the first place and thanks for that patch.

Issue #521, opened 2024-09-12
Error read input/include files in TeX files


lksj:
Using \input instruction in TeX file results in error
"No such file or directory".

```perl
po4a::tex: Cannot read from ./subtext1.tex
           : No such file or directory
 at /src/projects/po4a/po4a line 2039.
```
Existing tests do not cover it. All tests return OK.

#### Environment

Version po4a: latest commit of master branch

#### Files for test 

Create TeX-files using `\input` and/or `\include`:

For example:

test.tex

```TeX
text in test.tex

\input{subtext1}

another text in test.tex

\include{subtext2}
```

subtext1.tex

```TeX

text in subtext1.tex

another text in subtext1.tex

```

subtext2.tex

```TeX

text in subtext2.tex

another text in subtext2.tex

```

test.cfg

```config
[po4a_langs] en
[po4a_paths] test.pot $lang:test.$lang.po
[type: tex] test.tex $lang:test.$lang.tex

```
Run po4a:

```console
$PERLLIB=/path/to/po4a/lib /path/to/po4a/po4a test.cfg

```

You can see that the error occurs when using \input or \include or both, 
and does not occur when it is commented out.

### Reason of error

kpsewhich return full filename with EOL character at the end.

#### Proof

Run perl debugger

```console
$PERLLIB=/path/to/po4a/lib perl -d /path/to/po4a/po4a test.cfg
```

and command `c` (continue).

Output from perl debugger:

```perl
po4a::tex: Cannot read from ./subtext1.tex
           : No such file or directory
 at /src/projects/po4a/po4a line 1663.
 at /src/projects/po4a/lib/Locale/Po4a/TeX.pm line 962, <KPSEA> line 1.
	Locale::Po4a::TeX::read_file(Locale::Po4a::LaTeX=HASH(0x27e0300), "./subtext1.tex\x{a}", "subtext1", "UTF-8") called at /src/projects/po4a/lib/Locale/Po4a/TeX.pm line 1012
	Locale::Po4a::TeX::read_file(Locale::Po4a::LaTeX=HASH(0x27e0300), "/src/projects/po4a-test/test.tex", "test.tex", "") called at /src/projects/po4a/lib/Locale/Po4a/TeX.pm line 938
[...]

PR #522, opened 2024-09-12
TeX: Add format test for valid input command

lksj:
Fix mquinson/po4a#521

Existing tests check invalid input command (included file not exist) and not check valid input command (included file exist and must be include). This PR fix it.


mquinson:
Thanks for this very good set of bug report and PRs, that's really impressive.

PR #532, opened 2024-10-08
Remove DynaLoader

gemmaro:
This removes DynaLoader from some is-a relationships.

DynaLoader was introduced in 0a4e526ddd3a3a5e9cc92f1832833d21a37b314c, but was removed in dbbc6d2c763ce89a13af003871ecffa01a8edc30 long time ago.  So it should be safe to remove it now.


mquinson:
Thanks a lot, that's a good simplification!

PR #533, opened 2024-10-13
Use parent pragmas instead of base's @ISA variables

gemmaro:
This change uses the parent pragma instead of the `@ISA` variable provided by the base module.  It also fixes the return value of some modules.

This is not a user-facing change, so I haven't added it to the NEWS file.

Background: I was trying to replace Pod::Parser with Pod::Simple (though the try is far from complete), and found it might be better to make myself understand Perl's module system and clean up existing modules.


mquinson:
Hello,

thanks for this change. I don't have the energy to dig into this, and I'm very happy that you do these cleanups.

PR #538, opened 2024-10-29
Better accommodations for reading non-UTF-8 PO files

Fat-Zer:
Instead of reading the whole file and only then checking the charset, read just up to the first msgid/msgstr, check if it specifies a charset right away. If it does, imbue the file with the correct charset and read the rest.

In particular this:
- Avoid unnecessary warnings
- Avoid reading the file repeatedly

---------

The idea was previously mentioned [here](https://github.com/mquinson/po4a/pull/515#issuecomment-2440856593)


Fat-Zer:
I'm not sure why CI fails. Is it possible to somehow get it more verbose? e.g. to yank the po4a output from it?

Fat-Zer:
I'm not sure why CI fails. Is it possible to somehow get it more verbose? e.g. to yank the po4a output from it?
PS: locally tests run fine.

mquinson:
The logs read: 
```
Malformed test charset/po-iso8859 (PO file encoding: iso8859-1): no expected output. Please touch charset/po-iso8859/_output
# Looks like your test exited with 2 just after 17.
t/charset.t ........... 
# Subtest: master encoding: ascii (dstdir)
[...]
Dubious, test returned 2 (wstat 512, 0x200)
All 17 subtests passed 
```

Could you please add this file to your PR?

Thanks, 
Mt

Fat-Zer:
oh... that was a bit embracing >_<... I have completely overlooked the line and were too quickly jumping to confusion that it was some weird perl-encoding-handling issue.

fixed.

mquinson:
Don't be embarrassed. This test module isn't specially user-friendly, it's just that I'm used to it. 

Beyond, your code is much better than my previous attempt at reading non-UTF-8 PO files. Many thanks for that.

PR #539, opened 2024-11-03
add --target-lang option to po4a

jnavila:
This option should allow to transition to po4a from po4a-*, while retaining the same steps in makefiles

### DIFF
--- a/po4a
+++ b/po4a
@@ -8,7 +8,6 @@ eval 'exec perl -S $0 ${1+"$@"}'
 #
 # This program is free software; you can redistribute it and/or modify it
 # under the terms of GPL v2.0 or later (see COPYING).
-
 =encoding UTF-8
 
 =head1 NAME
@@ -292,6 +291,12 @@ This option can be used multiple times.
 Define a variable that will be expanded in the B<po4a> configuration file.
 Every occurrence of I<$(var)> will be replaced by I<value>.
 This option can be used multiple times.
+
+=item B<--target-lang> I<LANG>
+
+Update PO files and generate translated documents only for the
+specified language. This option can be used multiple times.
+
 
 =item B<--srcdir> I<SRCDIR>
 
@@ -780,6 +785,7 @@ sub get_options {
         "package-version"    => undef,
         "srcdir"             => undef,
         "destdir"            => undef,
+	"target-lang"        => [],
         "calldir"            => cwd()
     );
     Getopt::Long::config( 'bundling', 'no_getopt_compat', 'no_auto_abbrev' );
@@ -818,7 +824,8 @@ sub get_options {
         'previous'             => \$previous,
         'msgmerge-opt=s'       => \$msgmerge_opt,
         'srcdir=s'             => \$opts{"srcdir"},
-        'destdir=s'            => \$opts{"destdir"}
+        'destdir=s'            => \$opts{"destdir"},
+	'target-lang=s'        => \@{ $opts{"target-lang"} }
     ) or pod2usage();
 
     $opts{"verbose"} = scalar @verbose;
@@ -1042,11 +1049,27 @@ my $config_file = shift(@ARGV) || pod2usage();
 
 # Parse the config file
 my (@langs);
+my (@target_langs) = @{ $po4a_opts{"target-lang"}};
 my (%aliases);                # module aliases ([po4a_alias:...]
 my $pot_filename     = "";    # In split mode, this is the temp file name for the big POT
 my $pot_filename_cfg = "";    # This is never modified, useful in split mode to find the cfg content
 my (%po_filename);            # po_files: '$lang'=>'$path'
 my (%add_filename);           # Global addendum_files: '$lang'=>array of '$path'
+
+sub check_target_langs {
+    my ($nb, $line) = @_;
+    if ( scalar @target_langs ) {
+	# check that all languages are in the list of destination languages are in the list of languages
+	foreach my $lang (@target_langs) {
+	    die wrap_ref_mod( "$config_file:$nb", "", gettext("Specified destination language '%s' not found in the list of languages."), $lang )
+		unless grep { $_ eq $lang } @langs;
+	}
+	$line = "";
+	@langs = @target_langs;
+    }
+    return $line;
+}
+
 
 # The document hash table contains all the information about the files to translate.
 # Each key of the hash map a master file, ie a file to translate. For each such master file, we have a sub-hashmap containing:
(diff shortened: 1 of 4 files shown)


Fat-Zer:
Good, it's a long due one!

I have some suggestions:

- to reword the doc string, because as for now it sounds like it only affects update of the PO files. Maybe to something like:
    > Generate translated documents and update PO files only for the specified language. This option can be used multiple times.
- to warn if a language is not present rather than die. And die only if none of the specified languages are present
- let the test check that it affects PO file update as well
- maybe to rename the option to `--lang-only`, so it would be more consistent with `--translate-only`

jnavila:
> Good, it's a long due one!
> 
> I have some suggestions:
> 
> * to reword the doc string, because as for now it sounds like it only affects update of the PO files. Maybe to something like:
>   > Generate translated documents and update PO files only for the specified language. This option can be used multiple times.

OK, Just reversing the order: 

    Update PO files and generate translated documents only for the specified language. This option can be used multiple times.


> * to warn if a language is not present rather than die. And die only if none of the specified languages are present

I'm not fond of being merciful at the begining of the logs, while after, there can be tons of messages about translation levels not reaching the required threshold. In addition, I expect this option to be used once for each invocation (that's how I'm using it for each language in a Makefile), so this would not make a difference for the end users, while making the code slightly more complicated.

Do you have a specific use case in mind?

> * let the test check that it affects PO file update as well

I do not understand this request. Do you mean po file update instead of po file creation?

> * maybe to rename the option to `--lang-only`, so it would be more consistent with `--translate-only`

Even if my intention is that the option is used only one time in the command line, it can be used several times, and then, the option cannot be named '--*-only' for this case, because that would imply a unique use.

mquinson:
That's embarrassing because I thought that the main po4a script could already be used as a drop-in replacement. It was even the reason why I stepped into the deprecation of the other scripts... Thanks for taking care of this.

As for the comments of @Fat-Zer, I agree with @jnavila that dying is important when something goes seriously wrong, as people don't read the logs otherwise. An inconsistent argument is something serious to me.

For the parameter names, we could deprecate ``--translate-only`` to ``--no-update`` or something. It's rather easy to do: just accept both forms, document only the new name, and produce a warning that nobody reads when the old name is used. Having consistent naming schemas is important to me (but it's not part of this PR).

If I think about it, maybe ``--localized-lang`` is better than ``--dest-lang`` because ``dest``is used elsewhere in po4a to change the directory to which we should  while we use ``--localized`` in ``po4a-*`` scripts and ``--localized-charset`` in all scripts.

Fat-Zer:
> For the parameter names, we could deprecate --translate-only to --no-update or something.

`--translate-only [file]` implies `--no-update`, but it's intent is to limit which master-documents to translate. It doesn't make sense to anyhow replace it with `--no-update`. And to rename the option just because of this kinda-of-a misnomer IMHO will bring too few benefits for all the headache it will cause maintainers.


> If I think about it, maybe `--localized-lang` is better than `--dest-lang` because `dest`is used elsewhere in po4a to change the directory to which we should while we use `--localized` in `po4a-*` scripts and `--localized-charset` in all scripts.

This looks more like "set a lang of the localized file", which doesn't fit here... maybe just "--lang"?

jnavila:
Ok, there are already several propositions, so I'll add mine : `--target-lang`. To me, the `--localized-charset` option is really attached to the `po4a-*` utilities where files are managed one by one, whereas here, the option filters out the actions performed by po4a depending on criteria.

mquinson:
``--target-lang`` seems good to me.

Fat-Zer:
For me `--dest-lang` and `--target-lang` are both kinda ok...

PS: I still would prefer `--lang-only` more even with all the buts, but as I said I won't insist on that...

po4a
```diff
@@ -780,6 +785,7 @@ sub get_options {
         "package-version"    => undef,
         "srcdir"             => undef,
         "destdir"            => undef,
+	"target-lang"          => [],
```
Fat-Zer, line 788: A couple of extra spaces here
```suggestion
	"target-lang"        => [],
```

po4a
```diff
@@ -818,7 +824,8 @@ sub get_options {
         'previous'             => \$previous,
         'msgmerge-opt=s'       => \$msgmerge_opt,
         'srcdir=s'             => \$opts{"srcdir"},
-        'destdir=s'            => \$opts{"destdir"}
+        'destdir=s'            => \$opts{"destdir"},
+	'target-lang=s'          => \@{ $opts{"target-lang"} }
```
Fat-Zer, line 828: and here
```suggestion
	'target-lang=s'        => \@{ $opts{"target-lang"} }
```

po4a
```diff
@@ -1042,12 +1049,28 @@ my $config_file = shift(@ARGV) || pod2usage();
 
 # Parse the config file
 my (@langs);
+my (@destlangs) = @{ $po4a_opts{"target-lang"}};
```
Fat-Zer, line 1052: Probably should rename as well. And the same for `check_dest_langs` later.

t/cfg-args.t
```diff
@@ -80,6 +80,13 @@ push @tests, {
     'closed_path'    => 'cfg/*/',
     'expected_files' => 'single-uptodate.fr.po  single-uptodate.pot single-uptodate.man.fr.1',
 
+  },
+  {
+    'doc'            => '--target-lang',
+    'po4a.conf'      => 'cfg/args-global/po4a.conf',
+    'options'        => '--no-update --target-lang=de --target-lang=fr',
+    'expected_outfile' => '_output-target-lang',
+    'expected_files' => 'man.de.1 man.fr.1'
```
Fat-Zer, line 89: Also, alignment.

```suggestion
    'doc'              => '--target-lang',
    'po4a.conf'        => 'cfg/args-global/po4a.conf',
    'options'          => '--no-update --target-lang=de --target-lang=fr',
    'expected_outfile' => '_output-target-lang',
    'expected_files'   => 'man.de.1 man.fr.1'
```

PS: I also mentioned before about testing if it affects the translation generation as well (i.e. dropping `--no-update`), but after some reflection I'm not so sure whether it actually will be beneficial.

Fat-Zer (changes requested):

t/cfg-args.t
jnavila: Hello, thank you for this review.

For the test, I tried to remove the "--no-update" option but that led to lots of other errors which I did not fully understood, due to the test fixture not really adapted what needs to be achieved. I also think this would not be very beneficial.

Fat-Zer: Yeah... you are right... I've looked into it and it it would require adding a pot file, but it would mess up all the other tests of the same directory. It doesn't worth the effort.

Fat-Zer (approved):
LGTM, if I may

mquinson:
Thanks a lot for these good changes (and for the code review)

Issue #445, opened 2023-10-29
po4a modifes the msgid from man page


goeranu:
After having translated the manual page for `test.1` to Swedish, I noticed the translation's synopsis comes out incorrect. The underlying problem seems to be that `po4a` modifies the source, which fooled me into giving an incorrect translation.

More exactly, this line is part of the original manual page:
```
.B "[\& ]\&"
```
In the _pot_ file generated by `po4a` this shows up like this:
```
msgid "B<[\\& ]\">"
```
Notice that the final `\\&` has been changed into `\"`. That is, the _roff_ code for a zero-width mark has become a double quote as part of the string.

(Since this string doesn't actually contain anything to translate, I just copied it to the _msgstr_, resulting in the double quotes showing up in the final manual page. Now knowing about the bug, I can obviously work around it in this particular case.)

I attach an example with a very much stripped down manual page, and the result of running `po4a` on it.

[po4a-bug.tar.gz](https://github.com/mquinson/po4a/files/13198019/po4a-bug.tar.gz)

PR #540, opened 2024-11-12
man: fix po4a's incorrect handling of \& in doblequotes

Fat-Zer:
Namely do `s/\\&"/\\(dq/` only until after we handle all the quotes that might separate commands' arguments.

Closes: https://github.com/mquinson/po4a/issues/445

### DIFF
--- a/lib/Locale/Po4a/Man.pm
+++ b/lib/Locale/Po4a/Man.pm
@@ -1674,25 +1674,25 @@ sub splitargs {
     # anymore. Let's play safe.
     $arguments =~ s/\\ /$nbs/g;
     $arguments =~ s/^ +//;
-    $arguments =~ s/\\&"/\\(dq/g;
     $arguments =~ s/^ *//;
     while ( length $arguments ) {
+        my $a;
         if ( $arguments =~ s/^"((?:[^"]|"")*)"(?!") *// ) {
-            my $a = $1;
+            $a = $1;
             $a =~ s/""/"/g if defined $a;
-            push @args, $a;
         } elsif ( $arguments =~ s/^"((?:[^"]|"")*) *$// ) {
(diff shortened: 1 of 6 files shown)


lib/Locale/Po4a/Man.pm
```diff
@@ -1674,25 +1674,25 @@ sub splitargs {
...
             $a =~ s/""/"/g if defined $a;
-            push @args, $a;
         } elsif ( $arguments =~ s/^([^ ]+) *// ) {
-            push @args, $1;
+            $a = $1;
         } else {
             die wrap_ref_mod( $ref, "po4a::man", dgettext( "po4a", "Cannot parse command arguments: %s" ), $arguments );
         }
+        $a =~ s/\\&"/\\(dq/g if (defined $a);
+        push @args, $a;
```
mquinson, line 1695: Just reading the code, it seems to me that we could push `undef` to `@args` if `$a` is not defined. If that's impossible, then the test at the end of line 1694 is probably useless.

Fat-Zer: @mquinson , can you take a look at this one more time...

mquinson: You are perfectly right, I'm sorry for being both mistaken and slow...

Fat-Zer: @mquinson, not really wrong, I'm not actually sure whether there are any use in those checks either ;) As for slow... don't sweat it... I'm not the worst contestant for bad-time-management olympics myself...

mquinson: Thanks for your nice words, @Fat-Zer. I'm glad to have you with me on this project.

Issue #417, opened 2023-05-25
Please provide a bash completion file for the po4a script


Freed-Wu:
Can it provide completions for common shells? Like <https://git.savannah.gnu.org/cgit/parallel.git/tree/src/parallel#n1609>, or [perl-rename](https://github.com/pstray/rename). TIA!

jnavila:
po4a is a tool to translate documentation files, using gettext's po files. For programs, you should internationalize your source code by using the getext library. See https://www.gnu.org/software/gettext/ for further information.

Freed-Wu:
Sorry,` I means `po4a -<TAB>` can provide options of `po4a`, like `perl-rename` and `parallel`. :cry:

jnavila:
Ah, sorry for misunderstanding. There's no shell completion files in the project right now.

mquinson:
Ok, I misunderstood too. Let's reopen the issue then. Help is welcome here as I'm not an expert of these pesky bash completion files.

PR #542, opened 2024-11-13
[some help required] Add bash completions

Fat-Zer:
I've sketched some bash completions. Nothing extra fancy just command-line options and some a couple of options with predefined values.

To test you may `export BASH_COMPLETION_USER_DIR=/path/to/po4a/share/bash-completion/` and then play with `po4a -<TAB><TAB>` etc.

Note that the completions are using `<script> --help`, and since po4a has to process PODs to print it, the completions are lagging a bit (~200ms, not absolutely infuriating, but still noticeable and slightly annoying).

-----------

But I need some help with installation scripting, I have very little idea how to do it with perl. The files in `share/bash-completion/completions` are supposed to be installed to directory pointed by `pkg-config --variable=completionsdir bash-completion`. Usually it's `/usr/share/bash-completion/completions`.

Closes: https://github.com/mquinson/po4a/issues/417


gemmaro:
Can Getopt::Long::Complete (or alternatives) be used to support shell completion? (I haven't tried yet though)

Fat-Zer:
@gemmaro It would resolve the issue with speed, but I'm a bit hesitant of introducing the dependency. If we make it required, it would be somewhat annoying for people who don't need it, and if we make it optional, it'd be either impossible to keep trek of it from the bash's side or nontrivial for maintainers to install.

Maybe instead we can add a flag that will just list the flags without help?

gemmaro:
@Fat-Zer 

> It would resolve the issue with speed, but I'm a bit hesitant of introducing the dependency. If we make it required, it would be somewhat annoying for people who don't need it, and if we make it optional, it'd be either impossible to keep trek of it from the bash's side or nontrivial for maintainers to install.

Adding an optional completion package sounds like a good idea.
I think the optional dependency (in the `recommends` field in `Build.PL`) is great for packaging, since it doesn't break packages that are just version bumped.

> Maybe instead we can add a flag that will just list the flags without help?

I like the way the completion script has granular control (e.g. `--format` in the change.)  I wonder if it is possible to do this with listing flags.

Fat-Zer:
> Adding an optional completion package sounds like a good idea.

To be fair, I don't know about perl infrastructure enough to make this call neither it's my call to make anyway.

> I think the optional dependency (in the recommends field in Build.PL) is great for packaging, since it doesn't break packages that are just version bumped.

The problem with this is that you will have to have a bash_completion file around anyway. And how to communicate to the maintainer whether the file is needed? If we just won't install it, the opportunity for completion will be unnoticeable for end-users. If we install it, there will be a problem that if the Getopt::Long::Complete is absent and a user tries to complete something it will result in an incorrect call... I'm not sure how to resolve all this in a sane-ish manner...

> I wonder if it is possible to do this with listing flags.

You mean po4a's modules' options? Unfortunately as for now it won't be trivial as there are no way to query a module which options it supports.

mquinson:
Sorry for the delay. I am merging it as I think that it is the responsability of the downstream packagers to install these files properly.

Fat-Zer:
hmmm... the completions wasn't particularly good, but now hopefully nobody will notice them =))

PR #544, opened 2024-11-17
Do not touch a doc file if it has not changed

jnavila:
Writing the same file changes the timestamp, which is not good for use in Makefile dependencies. This commit avoids writing a doc file if the content has not changed from what is already in the file.

### DIFF
--- a/po4a
+++ b/po4a
@@ -719,6 +719,7 @@ use Pod::Usage qw(pod2usage);
 use File::Temp;
 use File::Basename;
 use File::Copy;
+use File::Compare;
 use File::Spec;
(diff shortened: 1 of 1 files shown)


Fat-Zer:
besides failing test: the `po4a` script uses spaces rather than tab-space mix for line shifts.

PS: the same problem was in #539, but I haven't noticed it before...

jnavila:
> besides failing test: the `po4a` script uses spaces rather than tab-space mix for line shifts.
> 
> PS: the same problem was in #539, but I haven't noticed it before...

OK, This is due to me mixing git and po4a development. Will reroll and fix the tabs.

PR #553, opened 2025-02-14
Fix test helper to exclude `--width` option from `po4a-updatepo` options

gemmaro:
This pull request addresses an issue with the test helper by ensuring that the options for `po4a-updatepo` do not include the `--width=<width>` option. This change may be related to commit faa7bc81720e7c3c5427d0bbde8c14998dd62934.


mquinson:
Thank you so much for this fix. I feel really ashamed that I did not find the time to fix my mess myself :bowing_man:

PR #559, opened 2025-03-28
Publish test container images

duck-rh:
Can be used to test po4a during development or when released.

The main motivation here is to simplify building the website since it requires built sources. This way website updates could be automated instead of the current manual process and then published via OpenShift.

The website deployment (aside from building and publishing the content) is done via Ansibel by RH OSPO and here is the corresponding change to switch to OpenShift: https://gitlab.com/osci/community-cage-infra-ansible/-/merge_requests/717

This requires a change in a role to deploy PHP on Openshift: https://gitlab.com/osci/ansible-role-openshift-apps/-/merge_requests/43
This role cannot use a lightweight web server in this case because the way Apache handles language file extensions (*.php.<lang>) in the original website could not be replicated (attempt with Caddy).

The `exit 1` in `01-build-pages.sh` when some language is unsupported for the footer was disabled to skip the error and be able to generate the rest. I'm not sure how this was done since the manual procedure is not ducumented and that may need to be revisited or the script updated.

The current draft bits are made using the image published in my PR branch and do not have webhooks as this requires access to this project I do not have but these will be changed later is accepted. We should be able to add webhooks for both the website change and the po4a sources changes.

Moreover the `01-build-pages.sh` script could be adapted to use po4a installed in PATH. The standard installation does this in /usr/local and this path could be added to PATH. That would allow to ship smaller images with /srv/po4a removed since most people would not need this.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -1,40 +1,61 @@
+---
 # This workflow will build po4a on linux using Module::Build
 name: Build on Linux CI
 
 on:
+  workflow_dispatch:
   push:
     branches:
       - '*'
-    tags-ignore:
+    tags:
       - '*'
   pull_request:
+    branches:
+      - '**'
 
 jobs:
   build:
 
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+      attestations: write
+      id-token: write
+
+    env:
+      REGISTRY: ghcr.io
+      IMAGE_NAME: ${{ github.repository }}
 
     steps:
-    - uses: actions/checkout@v3
-    - name: Install Debian dependencies
-      run: |
-        sudo apt update 
-        sudo apt install -y liblocale-gettext-perl libtext-wrapi18n-perl libunicode-linebreak-perl libpod-parser-perl libtest-pod-perl libyaml-tiny-perl libsyntax-keyword-try-perl
-        sudo apt install -y cpanminus gettext docbook-xml docbook-xsl docbook xsltproc 
-        sudo apt install -y texlive-binaries texlive-latex-base opensp libsgmls-perl
-    - name: Install CPAN dependencies
-      run: |
-        cpanm Locale::gettext
-        cpanm http://search.cpan.org/CPAN/authors/id/R/RA/RAAB/SGMLSpm-1.1.tar.gz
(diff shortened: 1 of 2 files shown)


duck-rh:
I have created an OpenShift project to test this deployment.

Our OpenShift instance use these IPs:
```
$ host router-default.apps.ospo-osci.z3b1.p1.openshiftapps.com
router-default.apps.ospo-osci.z3b1.p1.openshiftapps.com has address 52.41.106.217
router-default.apps.ospo-osci.z3b1.p1.openshiftapps.com has address 35.155.158.53
```

Using one of the IPs you can check the website using curl:
```
curl --resolve www.po4a.org:80:35.155.158.53 -H 'Accept-Language: fr' http://www.po4a.org/
```

duck-rh:
If switching to OpenShift then the apex cannot hold a CNAME to the OpenShift route but we can change the po4a.org IP to our magic web redirector and redirect HTTP/HTTPS to www.po4a.org that will hold the CNAME.

ArduanovDanil:
Great job! This project need official Docker image on Docker Hub! I created my own docker image and now can use po4a by simple command `docker run -it --rm -v $(pwd):/src arduanovdanil/po4a-fork:v0.73 po4a.cfg`. I wanted to make PR, but you did it first :)

duck-rh:
I changed the way the website is published on OpenShift to avoid having knowledge of the build rules on our end (despite most being in the build script that's already too much) and this also adds test to the website generation. Once built the website is made into an image that we publish on OpenShift (by composing images via multi-stage builds). You can find the new PR here: https://github.com/mquinson/po4a-website/pull/20

mquinson:
Hello, many thanks for this contribution!
I'm not completely sure of why this is still marked as a draft as LGTM. It also contains too much of docker-fu for me, but I really trust your expertise on this point and I'm willing to merge it.

As a side note, if we publish an official docker image, I think it should be based on alpine rather than on a fat ubuntu. I'll try to find the time to do so at some point, but this test image is already very good if it helps building the website in a better way.

Thanks again,

duck-rh:
@mquinson I forgot to mark it ready indeed. I'll need to adapt the other PRs to use the right image name and not my fork but let's get this in first.

PR #560, opened 2025-03-31
markdown: Prevent infinite loop with incomplete code block backticks

### LINKED ISSUE #558: misformated markdown: po4a enters an endless loop
A misformatted markdown file creates an endless loop in po4a

code blocks are supposed to start and finish with three `

when a code blocks stops with only two `, it creates an infinite loop in po4a

I'm only using the markdown option:

```
[po4a_alias:markdown] text opt:"--option markdown"
[po4a_langs] fr
[po4a_paths] translation_files/$master/$master.pot $lang:translation_files/$master/$lang.po
[type: markdown] src/conduit/ADMIN.md $lang:translated/$lang/src/conduit/ADMIN.md opt:"--keep 0" pot=conduit

```

we can see the source file there: https://github.com/YunoHost-Apps/conduit_ynh/blob/b7d035627df5093f257f2b11992aab86cd5c3fad/doc/ADMIN.md

expected: po4a should continue processing (it can ignore the file with a log error or generate a buggy content, whatever you prefer)

gemmaro:
This should close #558.

This update prevents potential infinite loops during parsing by implementing safeguards for unfinished code blocks.

Thank you.


mquinson:
Many thanks!

PR #561, opened 2025-03-31
Add Direnv config file for Perl development

gemmaro:
This introduces a dotenv file (`.env`) and a Direnv configuration file (`.envrc`) for Perl development. Additionally, the `.envrc` file includes optional setup for Guix integration.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,6 @@
+# This file is used by Direnv and it helps loading and upnloading environment
+# variables automatically when navigating directories.  For more information
+# about Direnv, visit the homepage: https://direnv.net/
+
+PERL5LIB="$PWD/lib:$PWD/t:$PERL5LIB"
+use guix --development po4a


mquinson:
Hello Gemmaro, thanks for this contribution.

what exactly is the usage of these files? The references to guix in `.envrc` speak for themselves, but I'm not sure I understand the purpose of the `.env` file. Who reads it, how is it used? Would you mind adding a line or two of comments to that file explaining that? a simple URL could be sufficient.

Thanks in advance and sorry for delaying the merge...

gemmaro:
Hello @mquinson, thank you for reviewing this.

The `.env` and `.envrc` files are intended for running tests on the console and editing Perl source code within this repository.  Specifically, the `.env` file is used to set Perl library paths for various dotenv-compatible tools, including—but not limited to—Direnv. 

For example, when using Direnv alongside Emacs' Direnv minor mode and Flymake (`perl-flymake`), opening a library file (e.g., `lib/Locale/Po4a/Pod.pm`) or a test file often triggers warnings about unknown modules, such as `Locale::Po4a::TransTractor`. However, if the `.env` file is loaded through `.envrc` and Direnv is allowed to process the `.envrc` file, Direnv sets the `PERL5LIB` environment variable automatically.  This ensures Flymake no longer complains about missing modules.

I agree that adding comments to clarify the purpose of these files is a good idea. I'll include a link to the Direnv homepage in the `.envrc` file for reference.

Additionally, I’m considering merging the `.env` file into the `.envrc` file to reduce clutter in the repository's root directory. Initially, I believed the `.env` file was useful beyond Direnv. However, I'm not sure about its usecases with other editors or plugins in fact (e.g., [Perl Navigator plugin](https://marketplace.visualstudio.com/items?itemName=bscan.perlnavigator) for VSCode might automatically detects local Perl modules).  If it’s later discovered that other tools can make use of the `.env` file, keeping it separate could enhance its usability across different environments.

Issue #551, opened 2025-01-28
VimHelp uses its own option to control the wrap column of the text rather than a generic one


Fat-Zer:
As for now vimhelp provides its own option to wrap the text [`textwidth`][1]. I would suggest to use the generic TransTractor's one `wrapcol` instead (which is controlled by `-w`/`--width` option of the scripts).

As for now the generic default is `76`, but the VimHelp's is `78`. Is this correct?

@gemmaro, any objections?

[1]: https://github.com/mquinson/po4a/blob/5d16870f490fbb0ca15c2bc11c0a4a5ffebf2d5a/lib/Locale/Po4a/VimHelp.pm#L378C9-L378C18

Fat-Zer:
Also there is codeblock [splitting][1] into several paragraphes. Is it actually desirable?

[1]: https://github.com/mquinson/po4a/blob/5d16870f490fbb0ca15c2bc11c0a4a5ffebf2d5a/lib/Locale/Po4a/VimHelp.pm#L279

gemmaro:
Hello @Fat-Zer,

> As for now vimhelp provides its own option to wrap the text [`textwidth`](https://github.com/mquinson/po4a/blob/5d16870f490fbb0ca15c2bc11c0a4a5ffebf2d5a/lib/Locale/Po4a/VimHelp.pm#L378C9-L378C18). I would suggest to use the generic TransTractor's one `wrapcol` instead (which is controlled by `-w`/`--width` option of the scripts).

Good catch!  That option is indeed more suitable for this purpose.  Since the VimHelp format support hasn't been released yet, removing the `textwidth` option seems reasonable.

> As for now the generic default is `76`, but the VimHelp's is `78`. Is this correct?

Yes, I'd like to set the default value to 78 for the VimHelp format, since the official Vim help files use this value (e.g. <https://github.com/vim/vim/blob/04cc8975930b7b2c5d6753d3eddf57dab2816518/runtime/doc/if_perl.txt#L307>).

> Also there is codeblock [splitting](https://github.com/mquinson/po4a/blob/5d16870f490fbb0ca15c2bc11c0a4a5ffebf2d5a/lib/Locale/Po4a/VimHelp.pm#L279) into several paragraphes. Is it actually desirable?

I prefer this approach since it results in smaller messages in the PO file.  However, it might be beneficial to have an option to change this behavior if combined code block messages are preferred.

gemmaro:
[...]

PR #570, opened 2025-04-27
Remove textwidth option of VimHelp module

gemmaro:
The `textwidth` option has been removed from the VimHelp module due to the complexity involved in achieving the desired behavior efficiently and effectively.  While initially intended to enhance functionality, its implementation presented challenges that made it impractical to retain under reasonable conditions.

Partially addresses #551.


mquinson:
Hello @gemmaro, thanks for this commit. I truly appreciate when we remove something that is not fully functionnal. I think that I should do more of this in po4a here and there, actually.

Thanks again,
Mt

PR #571, opened 2025-05-03
org: Fix to ensure the correct ordering of paragraphs and headings

gemmaro:
This fixes the ordering of paragraphs and headings to ensure correctness in the Org module.
It also addresses minor issues related to paragraph references.


mquinson:
nice catch, thanks.

PR #572, opened 2025-05-06
Add support for Emacs bug-reference mode

gemmaro:
This specifies a project-specific bug reference pattern and URL formats.
This allows bug tracker links to be highlightened, especially when the Emacs's bug-reference minor mode is enabled for the NEWS file, for instance.

Other project which has its own `.dir-locals.el` file: <https://git.savannah.gnu.org/cgit/guix.git/tree/.dir-locals.el>

po4a-Emacs-bug-reference-NEWS


mquinson:
Nice, thanks.

Issue #256, opened 2020-07-11
Consider migrating away from deprecated Pod::Parser


nanonyme:
https://metacpan.org/pod/Pod::Parser says `NOTE: This module is considered legacy; modern Perl releases (5.18 and higher) are going to remove Pod-Parser from core and use Pod-Simple for all things POD.` but this project marks it as a build dependency still.

mquinson:
This is still something on my TODO, unfortunately. Any help would be welcome.

mquinson:
As a side note, I guess that we've found a bug in Pod::Parser or something. In Locale::Po4a::Xml, we have `C<<? ... ?>>` in the source file, which is converted to `"<? ... ?">` (notice how the last quote is misplaced).

I don't know what happens here, but I suspect that Pod::Parser is failing to match the `<` of the input with the right `>`. Maybe switching to Pod::Simple will fix it?

PR #573, opened 2025-05-24
Add SimplePod module, which uses Simple::Pod to parse Pod documents

gemmaro:
Hello,

This introduces the `SimplePod` module, which utilizes the `Pod::Simple` module to parse POD documents.

Currently, the `Locale::Po4a::Pod` module relies on the `Pod::Parser` module for parsing, but this dependency has been deprecated.  The new `Locale::Po4a::SimplePod` module is designed as a drop-in replacement for the existing Pod module.

Since switching to the `SimplePod` module must be a significant change for translator, I propose the following transition steps:

1. The current Pod module issues a warning when parsing documents, informing users about the new `SimplePod` module and advising them to try it.
2. Refine the `SimplePod` module based on user feedback.
3. Replace the Pod module implementation with `SimplePod`.  The `SimplePod` name will remain available in the Po4a configuration file through `Locale::Po4a::Chooser`, with an added warning about its future removal.  It’s possible that a request may arise to retain the current Pod module under a different name, which seems reasonable.
4. Remove the `SimplePod` name from the Chooser module.
5. Mark the old Pod module as deprecated, and remove it from the distribution someday.

Releated issues:

* #256 
* https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=625924
* https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=962243

Thank you.

---

TODO list to be ready:

- [ ] Check diffs
  - [ ] Remove Pod::Parser from Build.PL.
  - [ ] Remove TODO in lib/Locale/Po4a/SimplePod/Parser.pm
- [ ] Refactoring
- [ ] More test cases
- [ ] Improve compatibilities


mquinson:
Whaou, this is a very good initiative, many thanks, Gemmaro! I'm really eager to see this great piece of code replacing the old parser. Thanks a lot

doc/po4a.7.pod
```diff
@@ -230,6 +230,13 @@ documentation).  Support for this format in po4a is still in its
 infancy.  If you find anything, please file a bug report or feature
 request.
 
+=item simplepod (very highly experimental parser)
```
mquinson, line 233: Is it still a very highly experimental parser? I'm just starting this review, but I kinda believe that it's stable now, even if it deserves more tests.

doc/po4a.7.pod
```diff
@@ -230,6 +230,13 @@ documentation).  Support for this format in po4a is still in its
...
+
+Similar to the previously mentioned I<pod>, this one adopts the new
+L<Pod::Simple> as its parser.  Since it is newly created, many bugs are
```
mquinson, line 236: Again, it's more "some bugs" rather than "many bugs" that are expected

lib/Locale/Po4a/SimplePod.pm
```diff
@@ -0,0 +1,147 @@
...
+
+The C<=begin> parameter section is no longer subject to translation.  In the
+current Pod module, it is included.  If there is a demand to include it in
+translation, please report it.
+
+=back
+
+=head1 STATUS OF THIS MODULE
+
+This module is still newly developed, so it is considerably less stable than
```
mquinson, line 124: considerably -> somewhat

mquinson (approved):
Ok, it seems to me that this new code has more tests than the previous one. Of course, it does not have the test of the external users yet, but I really don't see why we should delay its inclusion (beside of my work load these days -- sorry for the delay).

@gemmaro do you see anything we should do before changing the doc to mark this module as stable but young and merge it to the trunk?

gemmaro:
@mquinson Thank you for taking the time to review the pull request.  
I don't have any specific plans to make further changes at this time. 

However, I have some reservations about marking this module as stable.  I believe stability is more about how the module's behavior may change in the future rather than just the quantity of tests or quality metrics.  I think stability will increase with user feedback and through subsequent releases.

gemmaro:
I updated the 'STATUS OF THIS MODULE' section in SimplePod based on your suggestion. I also added the SimplePod module to the list of translation targets.

mquinson:
Thank you so much!

PR #588, opened 2025-07-04
vimhelp: add an option to split codeblocks

Fat-Zer:
This introduces an option to split codeblocks as discussed in [1]. The
default behaviour used to be to always split them. The new is to not,
unless requested via the option split_codeblocks.

Besides that the PR includes:
* [vimhelp] some refactoring
* [vimhelp] some new warnings
* [vimhelp] minor documentation mistake fixes
* small fixes to Testhelper
* small TransTractor fixes

[1]: https://github.com/mquinson/po4a/issues/551


mquinson:
shit! I said maybe one hour ago that we were in string freeze for the translators. Maybe we could rush this changes in before anyone reads my mail?

mquinson:
After reviewing the code, I have the feeling that it's a bit too large to rush it in. Maybe we should do 2 releases of po4a soon. One with what's already in master, and one with this PR and the fix to some of the bugs. We may hope that I'll find more time soon to fix them...

Fat-Zer:
Yep, sure... There is a one big commit and a couple of minor vaguely related fixes; I can split them off into dedicated PRs if that eases the review...

mquinson:
So this is it, the release is behind us. You can keep going on this one ;)
Thanks,

PR #594, opened 2025-07-12
Avoid implicit perl imports - make them explicit.

shlomif:
See:

https://perl-begin.org/tutorials/bad-elements/#non_explicitly_imported_symbols


mquinson:
Thanks a lot for your contribution :blush:

Issue #597, opened 2025-07-17
help2man - "bad argument to @<...>" - Translation Inlines Arguments


vindicatorr:
po4a version: 0.73
makeinfo version: 7.2

I was attempting to build the help2man repo source (https://salsa.debian.org/bod/help2man) (as opposed to the tar package), and encountered:
```
/usr/bin/vendor_perl/po4a-translate -f texinfo -m ./help2man.texi \
    -p po-texi/de.po -l help2man-de.texi -k 60
po4a-translate is deprecated. The unified po4a(1) program is more convenient and less error prone. Once configured, `po4a --no-update` can be used as a drop-in replacement to `po4a-translate`.
The TexInfo module of po4a is not ready for production use, and needs a new maintainer.
Please contact the po4a team if you want to help: send us patches, not bug reports.
(use -o no-warn to remove this message)
set -e; \
lang=de; \
locale=$(./build-aux//lookup-locale   $lang ./locales); \
/usr/bin/makeinfo --output=help2man-de.info.tmp --document-language="$locale" help2man-de.texi
help2man-de.texi:12: warning: @documentlanguage should only appear at the beginning of a line
help2man-de.texi:12: warning: @documentlanguage should not appear in @documentencoding
help2man-de.texi:12: warning: @paragraphindent should only appear at the beginning of a line
help2man-de.texi:12: warning: @paragraphindent should not appear in @documentlanguage
help2man-de.texi:12: bad argument to @documentlanguage: en @paragraphindent none
help2man-de.texi:12: bad argument to @documentencoding: UTF-8 @documentlanguage en @paragraphindent none
make: *** [Makefile:190: help2man-de.info] Error 1
```

I found that even though the `help2man.texi` correctly has the arguments on separate lines, the translation inlines them, resulting in `makeinfo` failing.
`$ less -N help2man-de.texi`:
```
      1 \input texinfo    @c -*-texinfo-*-
      2 @c ===========================================================================
      3 @c
      4 @c This file was generated with po4a. Translate the source file.
      5 @c
      6 @c ===========================================================================
      7 
      8 @c %**start of header
      9 @setfilename help2man.info
     10 @settitle @command{help2man}-Referenzhandbuch
     11 @c %**end of header
     12 @documentencoding UTF-8 @documentlanguage en @paragraphindent none
     13 
     14 @dircategory Software-Entwicklung
```

[...]

PR #598, opened 2025-07-18
Texinfo: add more missing commands

Fat-Zer:
Bug: https://github.com/mquinson/po4a/issues/597


mquinson:
Thanks a lot @vindicatorr I'll do that.

Issue #586, opened 2025-07-01
docbook: pot output has two spaces between sentences in para if dot is last character of the line


Wormnest:
For the GIMP manual I am testing out possible replacement of xml2po and started a small test of po2a. In doing so, I discovered the following issue.

Input xml is [https://gitlab.gnome.org/GNOME/gimp-help/-/blob/master/src/introduction/whats-new.xml?ref_type=heads](https://gitlab.gnome.org/GNOME/gimp-help/-/blob/master/src/introduction/whats-new.xml?ref_type=heads)

My test config:

```
[po4a_langs] fr de
[po4a_paths] pot/$master.pot $lang:po/$lang/$master.po

[type: docbook] src/introduction/whats-new.xml $lang:xml/$lang/introduction.xml pot=introduction
```

Command run: `po4a --verbose --destdir /home/Jacob/build/gimp-help po4a.cfg`

The output pot file has two spaces between `documentation.` and `However`:

```
#. type: Content of: <sect1><para>
#: src/introduction/whats-new.xml:14
msgid ""
"We realize that it would be nice to have localized changes of the latest "
"release that links to relevant pages in our documentation.  However, our "
"team just doesn't have enough volunteers to make that happen. We would "
"welcome more people helping out with updating or translating our "
"documentation, so don't hesitate to <ulink url=\"https://docs.gimp.org/"
"help.html\">contact us</ulink>."
msgstr "" 
```

mquinson:
Hello,

I see that there is two spaces between these words, but technically, there is 3 spaces between these words in the input file (the indentation):
```
  <para>
    We realize that it would be nice to have localized changes of the
    latest release that links to relevant pages in our documentation.
    However, our team just doesn't have enough volunteers to make that
    happen. We would welcome more people helping out with updating or
    translating our documentation, so don't hesitate to
    <ulink url="https://docs.gimp.org/help.html">contact us</ulink>.
  </para>
```

Is there a difference in the produced files? Are the indenting spaces handled differently from other spaces? I'm sorry I'm not aware of the specialties of the XML formatting toolchain.

If these spaces break something, I'll try to see where it comes from to fix it, but any helping hand is really welcome here.

Thanks for the info on what gets broken,
Mt

Wormnest:
> Is there a difference in the produced files? Are the indenting spaces handled differently from other spaces? I'm sorry I'm not aware of the specialties of the XML formatting toolchain.

[...]

PR #599, opened 2025-07-19
Pass --wrap-po option counterparts to gettext utils

Fat-Zer:
Pass appropriate `--width` or `--no-wrap` options in subsequent calls to msgmerge, msggrep, msgcat and mdginit according to `--wrap-po`.

Closes: https://github.com/mquinson/po4a/issues/586


Fat-Zer:
Sorry; wrong bug

PR #615, opened 2025-10-25
New parser for Texinfo

pertusus:
A new parser for Texinfo using the texinfoparser format name based on the GNU Texinfo libraries SWIG interface. The GNU Texinfo libraries SWIG interface is not released yet but should be in the next Texinfo release.

There is something that should be done for @documentlanguage, but I couldn't find a way to test a real language in the testsuite, so I left that for later.

There may be something specific to do with comments with TRANSLATIONS in them, so far they are treated as normal comments, but there is something in the other Texinfo implementation in POD and in tests about that.  Please tell me if there is something to do for that.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -21,6 +21,9 @@ AsciiDoc:
    (GitHub's #610 and #611) [Oliver Rahner]
  * Fix table header recognition pattern and context [Jean-Noël Avila]
 
+CommonMark:
+ * New module (GitHub's #643) [gemmaro].
+
 SimplePod:
  * Fix incorrectly emitted list item content twice in translated output
    (GitHub's #638 and #640) [gemmaro]


pertusus:
I tried to add support to the build system. I used todo for the tests, such that the missing dependency does not lead to tests failures.

Now there are many po/pot files modified when running ./Build. Should their changes be committed?

pertusus:
There is now a texinfo-swig Debian package for texinfo-7.3 that contains the SWIG interface needed for the texinfoparser parser.

As a side note, for this PR and also in the future after/if it is integrated in po4a, I intend to test with the latest released Texinfo version (currently 7.3) (in Debian testing) and the git development version. Both changes in the SWIG interface and in the tree representation could happen and impact the texinfoparser in po4a.

Also, I needed to push with force after rebasing, I expect there is no other possibility to keep up to date with the changes in the main repository and have the commits cleanly rebased on top of the main repository commits with conflicts fixed.

PR #619, opened 2025-12-09
asciidoc: match table delimiters more precisely

emteelb:
> A table is delimited by a vertical bar and three equal signs (|===).

see https://docs.asciidoctor.org/asciidoc/latest/tables/data-format/#default-table-syntax

### DIFF
--- a/lib/Locale/Po4a/AsciiDoc.pm
+++ b/lib/Locale/Po4a/AsciiDoc.pm
@@ -481,7 +481,7 @@ sub parse {
             $self->pushline( $line . "\n" );
         } elsif ( ( defined $self->{type} )
             and ( $self->{type} eq "Table" )
-            and ( $line !~ m/^\|===/ )
+            and ( $line !~ m/^\|={3,}$/ )
             and ( $self->{options}{"tablecells"} )
             and ( not defined $self->{disabletablecells} ) )
         {
@@ -1021,7 +1021,7 @@ sub parse {
             $line = "";
             undef $self->{bullet};
             undef $self->{indent};
-        } elsif ( $line =~ /^\|===/ ) {
+        } elsif ( $line =~ m/^\|={3,}$/ ) {
 
             # This is a table, treat it as a non-wrapped paragraph
             print STDERR "Found Table delimiter\n" if ( $debug{parse} );
--- a/t/fmt/asciidoc/DelimitedBlocks.adoc
+++ b/t/fmt/asciidoc/DelimitedBlocks.adoc
@@ -182,3 +182,14 @@ __________________________
 
 end of test
 --------------------------
+
+// table delimiter (`^\|={3,}$`) inside code block
+----
++---------------------------------------------------------------------+
+| lorem ipsum  | lorem ipsum  | lorem ipsum             | lorem ipsum |
+|=====================================================================|
+| lorem ipsum  | lorem ipsum  | lorem ipsum             | lorem ipsum |
+| lorem ipsum  | lorem ipsum  | lorem ipsum             | lorem ipsum |
+| lorem ipsum  | lorem ipsum  | lorem ipsum             | lorem ipsum |
++---------------------------------------------------------------------+
+----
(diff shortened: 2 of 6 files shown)


emteelb:
When creating translation files, the following code block contains a line that the current table delimiter in the Asciidoc plugin matches:

```
[source%unbreakable,bash]
----
+----------------------------------------------------------+
| Node   | NodeType  | Addresses                  | State  |
|==========================================================|
| kube-0 | SATELLITE | 172.16.145.70:3366 (PLAIN) | Online |
| kube-1 | SATELLITE | 172.16.126.69:3366 (PLAIN) | Online |
| kube-2 | SATELLITE | 172.16.79.133:3366 (PLAIN) | Online |
+----------------------------------------------------------+
----
```

This causes mistaken labeling/commenting in the `.pot` file:

```
#. type: Table
#: en/linstor-kubernetes-quick-start.adoc:286
#, no-wrap
msgid ""
"+----------------------------------------------------------+\n"
"| Node   | NodeType  | Addresses                  | State  |\n"
"| kube-0 | SATELLITE | 172.16.145.70:3366 (PLAIN) | Online |\n"
"| kube-1 | SATELLITE | 172.16.126.69:3366 (PLAIN) | Online |\n"
"| kube-2 | SATELLITE | 172.16.79.133:3366 (PLAIN) | Online |\n"
"+----------------------------------------------------------+\n"
"----\n"
msgstr ""
```

The stanza here should be labelled/commented as a "delimited block" but is instead labelled as a "Table". And because in this case, there is no closing table "delimiter" in the stanza, subsequent translation stanzas are all labelled "Table" mistakenly.

emteelb:
Thanks for taking the time to comment on this PR, @jnavila . I can work to create a test case here.

I'm confused about you citing asciidoc-py docs when that seems to be a Python project and po4a, at least as far as the GitHub languages summary reports contains no Python code.

Though from the current AsciiDoc maintainer, [Asciidoctor docs](https://docs.asciidoctor.org/asciidoctor/latest/migrate/asciidoc-py/), maybe the reason is simply that AsciiDoc.py was the original AsciiDoc language maintainer?

> AsciiDoc.py is the original and legacy Python-based processor for the AsciiDoc language. It has been superseded by Asciidoctor.

Later in that page I read:

> The Asciidoctor processor is a drop-in replacement for AsciiDoc.py. You can replace the call to AsciiDoc.py (asciidoc) with the equivalent call to Asciidoctor (asciidoctor).

If the po4a project does not use the AsciiDoc.py parser itself, should not po4a use the syntax from the current AsciiDoc maintainer (Asciidoctor)?

Or is there another relationship between po4a and the legacy AsciiDoc.py project?

If there is a reason to preserve the legacy "at least 3 equal signs" AsciiDoc.py syntax for table delimiters rather than the current "three equal signs" then the alternative change you showed that would fix my issue seems fine to me.

And yes, if possible, looking into the larger issue of why a delimiter was interpreted within a "code block" would be great.

Thanks for the help and direction here.

jnavila:
see https://gitlab.eclipse.org/eclipse/asciidoc-lang/asciidoc-lang/-/merge_requests/53 for a statement from the maintainer of asciidoctor

t/fmt/asciidoc/DelimitedBlocks.adoc
```diff
@@ -182,3 +182,14 @@ __________________________
 
 end of test
 --------------------------
+
```
jnavila: You cannot just add the delimited block to the `.adoc` file and expect the tests to pass... This change needs to be sided by the changes in the `pot`, `po`, `trans` and `norm` files.

lib/Locale/Po4a/AsciiDoc.pm
```diff
@@ -481,7 +481,7 @@ sub parse {
             $self->pushline( $line . "\n" );
         } elsif ( ( defined $self->{type} )
             and ( $self->{type} eq "Table" )
-            and ( $line !~ m/^\|===/ )
+            and ( $line !~ m/^\|={3,}$ )
```
jnavila: This part is ok, as discussed. But not the one about source section, which the test is targeting.

jnavila:
> Do I need to close this pull request then in favor of #622 then or are you able to do that as a maintainer?

I credited you for finding this bug. I'll push my versions if you do not mind.

Issue #253, opened 2020-07-07
Feature request: gettext context


paususe:
When po4a-gettextize creates a POT file from an AsciiDoc source, it only adds `msgid` and `msgstr` attributes:
```
#. type: Plain text
#: ./modules/installation/pages/install-intro.adoc:16
msgid ""
"{susemgr} Server and Proxy are available from {sle} as base products, and "
"can be installed with the Unified Installer.  Installing with the Unified "
"Installer is the default method of installation."
msgstr ""
"{susemgr} Server y Proxy están disponibles desde {sle} como productos base, y "
"pueden ser instalados con el Instalador Unificado. Instalar con el Instalador "
"Unificado es el método de instalación por defecto."
```

Gettext supports the `msgctxt` attribute to provide context to translators and disambiguate when the sentence is not clear enough, which would be useful for documentation written in AsciiDoc.

Since AsciiDoc supports comments, this could be implemented with a special type of comment, e. g:
```
// TRANSLATORS: <whatever goes into the msgctxt>
```

Useful links:

* gettext contexts: https://www.gnu.org/software/gettext/manual/html_node/Contexts.html
* msgctxt rationale and example: http://pology.nedohodnik.net/doc/user/en_US/ch-poformat.html#sec-podisamb
* AsciiDoc comment syntax: https://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#comments
* Uyuni documentation PR, were we are planning to use po4a to convert AsciiDoc sources to gettext files and provide them to translators: https://github.com/uyuni-project/uyuni-docs/pull/325
* Uyuni Project in the openSUSE WebLate, which we will use to translate the docs (and the WebUI and the command-line tools :-) ) : https://l10n.opensuse.org/projects/uyuni/
* WebLate supports `msgctxt`: https://docs.weblate.org/en/latest/user/translating.html

ciampix:
This would be GREAT!!!

mquinson:
Hello,

we discussed it on IRC and I told you that this was possible for some formats of po4a but not in asciidoc yet. But I was wrong. This is already handled in asciidoc, and it's even tested: 
From this [adoc file](https://raw.githubusercontent.com/mquinson/po4a/master/t/fmt/asciidoc/Comments.adoc), po4a produces this [POT file](https://raw.githubusercontent.com/mquinson/po4a/master/t/fmt/asciidoc/Comments.pot). As you can see, the comments are properly added as comments into the POT file.

[...]

PR #627, opened 2026-01-11
Support gettext msgctxt

gemmaro:
Hello,

This change adds support for gettext `msgctxt` in po4a and can be used across all formats.

The motivation for this feature is explained in the comment at https://github.com/mquinson/po4a/issues/253#issuecomment-3697916562. In short, it is useful when `msgid` values tend to be short and need to be translated differently depending on context.

The pull request contains many commits that can be roughly divided into several phases:

1. Enhance encapsulation of the Po module and its usage by other modules
2. Modify the Po instance data structure by inserting a `msgctxt` attribute layer between `msgid` and other message fields (e.g., `msgstr`)
3. Implement `msgctxt` support in the Po module
4. Add an option to accept context in the TransTractor module (which automatically enables the option to be used in po4a config files)
5. Add `--context-module` option for po4a
6. Add tests for the existing/updated implementation

Note: I initially added the `context-module` option to be used with `--option` (e.g., `opt: "--option context-module=MODULE_NAME"`), but replaced it with a dedicated `--context-module` option since `--option` is intended for additional parameters specific to individual format modules.

Thank you.

---

<details>

<summary>TODO to be ready for review:</summary>

- [x] Write document comments for updated methods (subroutines) in Po module
- [x] Write context option usage.  Maybe more detailed description should be provided in po4a(1 or 7) manpage.
  - [x] tips: note about string escape (unicode normalization), which might be confusing such as `’` and `\x{2019}`
- [x] Drop unrelated commits to this topic
- [x] Add more tests, especially unit tests for existing functionalities
- [x] Cleanup unit tests for Po with subtest
- [x] Enable to pass additional criteria to generate msgctxt by each format modules?
- [x] Update news
- [x] Update this comment
- [x] Add more useful parameters I found to be useful along translating
- [x] ~~bug fix: do not let updatepo include msgctxt in `#|` comments~~ <- it wasn't a bug
- [x] change: Add `--context-module=MODULE_NAME` option instead of utilizing `--option`.

</details>

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -9,6 +9,9 @@ __   __/ _ \|___  | ___|
  \ V /| |_| | / /  ___) |
   \_/  \___(_)_/  |____/          (unreleased)
 
+General:
+ * Support gettext msgctxt (GitHub's #627) [gemmaro]
+
 AsciiDoc:
  * Fix line numbers and order when include macro is inside of lists
    (GitHub's #610 and #611) [Oliver Rahner]


lib/Locale/Po4a/Po.pm
```diff
@@ -351,46 +351,50 @@ sub read {
     }
 
     my $is_charset_detected;
+
     # Detect the charset
-    if ( $pofile =~ /^msgid ""\s*$/m &&
-         $pofile =~ /^msgstr ""\s*$/m &&
-         $pofile =~ /charset=(.*?)[\s\\]/
-    ) {
+    if (   $pofile =~ /^msgid ""\s*$/m
```
jnavila: are you using a code formatter for this? In this case, it would be good to share it, so that we can all settle on a common formatting.

mquinson:
Friends, I'm sorry, I really struggle to find some time to devote to po4a. It's high in my TODO but never on top unfortunately. I hope to find some time this month :crossed_fingers:

jnavila:
43 commits in one PR ! That's not reviewable !

PR #641, opened 2026-02-18
Minor refactoring: simpify debug/verbose option handling

gemmaro:
This is a minor refactoring of `debug` and `verbose` options, allowing format modules to more easily ignore them when overriding the `initialize` subroutine.  Additionally the `debug` subroutine is utilized in some places instead of direct attribute accessings.


mquinson:
You're right, it looks much better this way. Thanks

PR #642, opened 2026-02-23
Minor refactoring: Extract YAML front-matter handling into its own module

gemmaro:
This change extracts the YAML front-matter processing logic from the `Text` module (with `markdown` option) into a dedicated module.
This separation makes it easier to introduce a `CommonMark` format module (as proposed in #639).

<details><summary>TODOs</summary>

TODOs before removing draft mark:

- [x] Add minimal POD documentation for the new module
- [x] Improve option passing from `$self->{options}{yfm_...}` to keyword arguments

</details>

### DIFF
--- a/lib/Locale/Po4a/Text.pm
+++ b/lib/Locale/Po4a/Text.pm
@@ -50,10 +50,9 @@ use 5.16.0;
 use strict;
 use warnings;
 
-use parent qw(Locale::Po4a::TransTractor);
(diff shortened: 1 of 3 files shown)


mquinson:
Excellent. This is something I wanted to do but never found the time to. Thanks in advance for that!

mquinson:
Thanks a lot!

PR #643, opened 2026-02-28
New module for CommonMark documents

gemmaro:
Hello,

This adds a new module for the CommonMark and closes #639.  This is also a continuation from #642 to fix a minor point.
Note that this is not intended to replace the `Text` module with the `markdown` option since the CommonMark doen't have additional features such as footnotes or tables.

Thank you,

---

<details><summary>TODOs before mark as ready:</summary>

- [x] Add tests
- [x] Add mentions to this module in the Chooser module and others (such as readme).
- [x] Remove unrelated changes (?)
- [x] Update issue number in the news
- [x] Fix Docker image prerequisites

</details>


doc/po4a.7.pod
```diff
@@ -237,6 +237,14 @@ L<Pod::Simple> as its parser.  Since it is newly created, some bugs are
 expected.  If you notice any strange behavior, please let us know.
 Eventually, this module will replace I<pod>.
 
+=item CommonMark (very highly experimental parser)
```
mquinson: Such sentences tend to keep in the code once integrated. I'd prefer if you could rewrite it as "newly introduced in 2026" which means "not really tested" in 2026 and 2027, but tend to change its meaning as the time passes.

mquinson:
I missed this PR, sorry. I feel it's good to merge, isn't it ? I only have the minor rewording that I added as a comment, please. Likewise, I feel that you are very harsh with your code. You say "very experimental code" where you could also say without lying "no known issue". It's true it's untested, but it should work reasonably, I guess. We already had such discussions in the past so I don't expect you to dramatically change your text as I suggest, and that's OK. I just wanted to say once again that you should be less critical to your own work :)

Thanks for this. I'll try to hit the merge button once you do the little rewrite to fix my comment.

mquinson:
By the way, I said I wanted to give you the write access to the repository. I still think about it. My only concern is that you should not change asciidoc before talking to @jnavila. I consider that he is in charge on that module and would not want bad interactions in the git, please.

doc/po4a.7.pod
gemmaro: Understood. I've applied your suggestion.

mquinson:
Thanks a lot, it's merged :)

gemmaro:
@mquinson Thank you for the review and for your kind words about the implementation.

Regarding the write access to the repository, thank you for considering it. I am aware that jnavila is in charge of the asciidoc module, and I will make sure to consult with him before making any changes to that area.
I should mention that I may be somewhat less responsive for a while due to personal circumstances, but I would be happy to access the access.

Thanks again for your patience and support!

mquinson:
Ok, I granted you the write access, but please don't see this as a burden. Only contribute when you can and when you find it interesting or fun, or something. We have enough burden in our lives to not add artificial ones.

Issue #73, opened 2018-02-06
Adding PO file headers to translated documents


mquinson:
First reported on Alioth on 27/09/2004 by me:
```
the po-debiandoc tool writes the PO-file header in the generated
SGML file (enclosed in <!-- ... -->, of course), so that these
useful informations are contained in this file.
It would be really nice if po4a-translate did something similar.


(Reported by Denis on the mailing list)
```

mquinson:
This should probably be fixed in each transtractor separately, as the main infrastructure does not know the comments syntax in the current format.

gemmaro:
Now the pull request #646, which implements `header_po()`, has been merged, we are left with updating each modules.

PR #644, opened 2026-03-19
Adding PO file headers to translated documents 

yugoo90:
Added header subroutines to the Po and TransTractor modules. Also added a test case for the header subroutine in the Po.pm module.


gemmaro:
I have two points; 1) what is the purpose of adding the header subroutines, and 2) it would be nice if you could clean up the commit history.

yugoo90:
Hi @gemmaro,

Thank you for responding to me. I am new to open source contribution so I apologize for not cleaning up the commit history. To answer your first point. I added the po header subroutines so that other modules can use the po headers in their own way. Also it was recommended to add header subroutines to the TransTractor and Po modules specifically. (https://github.com/mquinson/po4a/issues/73).

lib/Locale/Po4a/TransTractor.pm
```diff
@@ -1373,6 +1367,12 @@ sub handle_yaml {
     }
 }
 
+# return the header of the PO file
+sub headerPO {
```
gemmaro: The subroutine name should use snake\_case (`header_po`), or all lowercase with no separators (`headerpo`) to match the naming style of the other subroutines.
Also, it might be helpful to have POD documentation for this subroutine. This is because this can be used by format modules (like `Locale::Po4a::Text`), so format module writers find this utility.

gemmaro (commented):
Hi.  Thank you for the clarification.  I understand the first point--purpose of this PR--and it looks good.  For the second point, if you're interested, you can try to clean up the history with [git-rebase](https://git-scm.com/docs/git-rebase).  I recommend this article, ["git rebase in depth"](https://git-rebase.io/), in case you haven't done this kind of thing before.

yugoo90:
Hi @gemmaro ,

I have opened a new pull request with the recommended changes. I will update the SGML module to use the po header along with tests for this.

PR #650, opened 2026-04-05
More explicit information on creating a PR with forking, branch, follow upstream, rebase and force push

pertusus:
Expand on the creation of PR, tracking upstream and the need for force push, as comprehensive description on that kind of workflow is not available anywhere, and somebody new would have to patch up diverse sources of documentation to get it right


mquinson:
Excellent, thanks a lot!

PR #653, opened 2026-05-24
asciidoc: preserve user styles in delimited blocks

jnavila:
Keep block fence tracking separate from the translation type, preserve user-defined verbatim-like styles for block content, and update StyleMacro expectations accordingly.


gemmaro (approved):
Sounds good to me.  This should be also helpul when gettextizing existing documents that use custom block styles.

(I found #658, but since it is an existing issue, I don't think it should block this change.)

jnavila:
Thanks for the review. This is really a workaround, because I am reluctant to dig deeper in this code and implement a proper context stack.

jnavila:
> This should be also helpul when gettextizing existing documents that use custom block styles.

That's exactly my use-case: the git manpages have a "synopsis" style which is verbatim.

PR #664, opened 2026-09-21
add PO header to SGML output

yugoo90:
- The SGML backend now includes PO metadata as an SGML comment. 
- The header is added only when a PO file was supplied.
- The two SGML expected-output fixtures were updated. 
- The SGML tests and the broader relevant test suite pass.


mquinson:
Hello @yugoo90, thanks for your contribution. 

I'd appreciate if you could modify the header of the relevant PO file so that the included comment in the SGML become more informative. I mean, since it's the usual placeholder in the PO file, it's not clear that the content of the SGML comment is really extracted and not just the usual thing.

Just change "SOME DESCRIPTIVE TITLE" to a more descriptive text, please. No need to be long or perfect, just ensure it's specific to that file.

Thanks again for your work,
Mt

